PR #1, opened 2020-01-22
feat: First PoC of headless Hathor Wallet

msbrogli:


api-docs.js
```diff
@@ -0,0 +1,160 @@
...
+                    value: {"statusCode":3,"statusMessage":"Ready","network":"mainnet","serverUrl":"https://node2.mainnet.hathor.network/v1a/","serverInfo":{"version":"0.29.0","network":"mainnet","min_weight":14,"min_tx_weight":14,"min_tx_weight_coefficient":1.6,"min_tx_weight_k":100,"token_deposit_percentage":0.01}}
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    '/0/balance': {
```
pedroferreira1: Why `/0/`? Is it the token UID?

package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "hathor-wallet-headless",
+  "version": "1.0.0",
+  "description": "Hathor Wallet Headless, i.e., without graphical user interface",
+  "main": "index.js",
+  "dependencies": {
+    "@babel/cli": "^7.8.3",
+    "@babel/core": "^7.8.3",
+    "@babel/node": "^7.8.3",
+    "@babel/preset-env": "^7.8.3",
+    "@hathor/wallet-lib": "file:../hathor-wallet-lib",
```
pedroferreira1: We should add the version of the package here

store.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export default class MemoryStore {
```
pedroferreira1: No need to create it, we already have on the wallet lib a `MemoryStore` that can be used.

wallet.js
```diff
@@ -0,0 +1,359 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import hathorLib from '@hathor/wallet-lib';
+import MemoryStore from './store';
```
pedroferreira1: You can import directly here the `MemoryStore` from wallet-lib

index.js
```diff
@@ -0,0 +1,111 @@
...
+  wallet.sendTransaction(address, value, tokenUid).then((response) => {
+    res.send(response);
+  }, (error) => {
+    res.send({success: false, error});
+  });
+});
+
+app.use(express.json());
+app.use(express.urlencoded({ extended: true }));
+app.use('/0', walletRouter);
```
pedroferreira1: Now I see the `/0/` from docs here but why are you using this?

wallet.js
```diff
@@ -0,0 +1,359 @@
...
+
+    const updateOutputs = (val) => {
+      const historyTxs = 'historyTransactions' in walletData ? walletData.historyTransactions : {};
+      for (const input of txData.inputs) {
+        historyTxs[input.tx_id]['outputs'][input.index]['spent_by'] = val;
+      }
+    }
+
+    // XXX Setting input as spent by, so it won't be chosen again
+    updateOutputs('2');
```
pedroferreira1: I know it's a PoC and we could continue with this here in this PR but we should handle this properly soon.

wallet.js
```diff
@@ -0,0 +1,359 @@
...
+        reject(error);
+      });
+    });
+    return promise;
+  }
+
+  /**
+   * Close the connections and stop emitting events.
+   **/
+  stop() {
```
pedroferreira1: Maybe this method should call the `cleanWallet` from `hathor-wallet-lib`.

We also need to unsubscribe to the addresses from the server, the WebSocketHandler also do some more cleanings. I think it's worth checking what's done there.

We just need to define what we want with this `stop` method. If the ideia is that we must start all over again (maybe with a new seed and server) after the stop, we must do the full clean.

api-docs.js
msbrogli: I though that it may have multiple wallets. Maybe using a different derivation in the HD path, or instantiating more wallets. It would be nice if the wallets could share one connection.

I also used it to put a middleware that checks if the wallet is ready. If it is not, it returns an error.

package.json
msbrogli: Exactly. I'm just waiting for the release of the new version.

index.js
msbrogli: I explained in the other comment.

store.js
msbrogli: Fixed. Thanks!

wallet.js
msbrogli: Fixed. Thanks!

wallet.js
msbrogli: As we close the websocket connection, we don't have to unsubscribe the addresses, right?

But I agree that we must a better definition of what this method should do. I'm not sure if we should run `cleanWallet`, but maybe we should add a `resetAll()` method to do a full clean up. What do you think?

wallet.js
msbrogli: You're right. We must also update the input status when a transaction is updated.

package.json
pedroferreira1: `0.9.0` released

wallet.js
pedroferreira1: Makes sense about the connection closed. But we should check if there is nothing else we should do.

wallet.js
pedroferreira1: Yes, but again...for a PoC I am ok going with it

api-docs.js
pedroferreira1: Humm understand. I think it's ok

wallet.js
```diff
@@ -0,0 +1,354 @@
...
+
+    hathorLib.network.setNetwork(network);
+    this.network = network;
+
+    this.state = Wallet.CLOSED;
+    this.serverInfo = null;
+
+    this.onConnectionChange = this.onConnectionChange.bind(this);
+    this.handleWebsocketMsg = this.handleWebsocketMsg.bind(this);
+    this.onAddressesLoaded = this.onAddressesLoaded.bind(this);
```
obiyankenobi: This is not doing anything. Is it just for future use?

wallet.js
```diff
@@ -0,0 +1,354 @@
...
+    const newTx = wsData.history;
+
+    // TODO we also have to update some wallet lib data? Lib should do it by itself
+    const walletData = hathorLib.wallet.getWalletData();
+    const historyTransactions = 'historyTransactions' in walletData ? walletData.historyTransactions : {};
+    const allTokens = 'allTokens' in walletData ? walletData.allTokens : [];
+
+    let isNewTx = false;
+    if (newTx.tx_id in historyTransactions) {
+      isNewTx = true;
```
obiyankenobi: If the tx is in `historyTransactions` it's marked as new? Shouldn't it be the opposite?

wallet.js
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    const walletData = hathorLib.wallet.getWalletData();
+    const historyTxs = 'historyTransactions' in walletData ? walletData.historyTransactions : {};
+
+    // First I need an array with all tokens
+    const allTokens = [];
+    for (const d of data) {
+      allTokens.push(d.token);
+    }
```
obiyankenobi: There would be repeated tokens in `allTokens` if different items of `data` have the same token.

package.json
msbrogli: Fixed. Thanks!

wallet.js
msbrogli: I marked it as a TODO item. So, we can review it when adding it to the hathor-wallet-lib. What do think? Can I mark it as solved?

wallet.js
msbrogli: I marked it as a TODO item. So, we can review it when adding it to the hathor-wallet-lib. What do think? Can I mark it as solved?

wallet.js
msbrogli: You're right. Fixed. Thanks!

wallet.js
msbrogli: Now it is propagating the event. Thanks!

wallet.js
msbrogli: Fixed. Thanks!

README.md
```diff
@@ -0,0 +1,184 @@
+# Hathor Wallet Headless
+
+This is a headless Hathor Wallet, i.e., without a graphical user interface. It is controlled by its HTTP API.
+
+It may be used both by mining pools and exchanges.
+
+## How to run?
+
+Copy `config.js.template` to `config.js`, and then fulfill the variables. Finally, run `npm start`.
```
jansegre: fullfill -> fill in

api-docs.js
```diff
@@ -0,0 +1,176 @@
+
+const apiDoc = {
+  openapi: '3.0.0',
+  servers: [
+    { url: "http://localhost:8000" }
+  ],
+  info: {
+    title: 'Headless Hathor Wallet API',
+    description: 'This wallet is fully controlled through an HTTP API.',
+    version: '1.0.0',
```
jansegre: Nitpick: I'd try to use `0.1.0` for anything that just started.

wallet.js
pedroferreira1: Ok for me.

wallet.js
pedroferreira1: Ok

api-docs.js
pedroferreira1: I agree, actually all our repositories have version `0.X.Y`

README.md
msbrogli: Fixed. Thanks!

api-docs.js
msbrogli: I agree. Fixed. Thanks!

1amdpreist:
Codex

PR #3, opened 2020-01-28
feat(auth): Add support for a simple API Key verification

msbrogli:
Maybe it closes #3, but we need to discuss it further.

This implementation follows the [Swagger Specification for API Keys](https://swagger.io/docs/specification/authentication/api-keys/).

Before merging:

- [x] Change to branch `dev` after `feat/poc` is merged.

### DIFF
--- a/config.js.template
+++ b/config.js.template
@@ -3,6 +3,9 @@ module.exports = {
   http_bind_address: 'localhost',
   http_port: 8000,
 
+  // Uncomment the following line to enable API Key Auth.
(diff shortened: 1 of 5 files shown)


config.js.template
```diff
@@ -2,6 +2,7 @@ module.exports = {
   // HTTP API Settings
   http_bind_address: 'localhost',
   http_port: 8000,
+  http_api_key: null,
```
pedroferreira1: Should we add a comment here saying it's optional?

msbrogli: Fixed. Thanks!

PR #5, opened 2020-01-31
refactor: using wallet class from lib

pedroferreira1:
- [x] Still need to update the lib version


index.js
```diff
@@ -9,20 +9,12 @@ import express from 'express';
 import hathorLib from '@hathor/wallet-lib';
```
msbrogli: I think we should change to `import { Wallet } from @hathor/wallet-lib`.

pedroferreira1: Done!

PR #8, opened 2020-02-24
feat: create new api to generate a new seed

pedroferreira1:


api-docs.js
```diff
@@ -55,6 +55,25 @@ const apiDoc = {
         },
       },
     },
+    '/new-seed': {
```
msbrogli: It is not secure to return the seed, and there is no way to use this seed in the wallet. I don't understand why we would add such method.

pedroferreira1:
This PR is not required anymore

PR #10, opened 2020-03-04
feat: adding support for multiple wallets

pedroferreira1:


README.md
```diff
@@ -12,18 +12,59 @@ Copy `config.js.template` to `config.js`, and then fill in the variables. Finall
...
+Create and start a new wallet.
+
+Parameters:
+
+`key`: key reference of this wallet in the object of created wallets. Used to say which wallet each request is directed to.
+`passphrase`: Optional parameter to start the wallet with a passphrase. If not sent we use empty string.
+`seedKey`: Optional parameter to define which seed (from the object seeds in the config file) will be used to generate the wallet. If not sent we use 'default'.
+
+```bash
+$ curl -X POST --data "key=key"  --data "passphrase=123" --data "seedKey=default" http://localhost:8000/start
```
obiyankenobi: 2 spaces after `"key=key"`.

README.md
```diff
@@ -12,18 +12,59 @@ Copy `config.js.template` to `config.js`, and then fill in the variables. Finall
...
 
+### Start a wallet
+
+Create and start a new wallet.
+
+Parameters:
+
+`key`: key reference of this wallet in the object of created wallets. Used to say which wallet each request is directed to.
+`passphrase`: Optional parameter to start the wallet with a passphrase. If not sent we use empty string.
+`seedKey`: Optional parameter to define which seed (from the object seeds in the config file) will be used to generate the wallet. If not sent we use 'default'.
```
obiyankenobi: Missing line breaks between the 3 lines. You have to add 2 spaces after the lines.

README.md
```diff
@@ -190,3 +231,24 @@ $ curl http://localhost:8000/0/tx-history
   }
 }
 ```
+
+### Stop a wallet
+
+Stop the wallet and its connections and remove it from the store.
```
obiyankenobi: What's the store?

pedroferreira1: It's the wallets object the have all the wallet instances. Maybe it's not clear

README.md
```diff
@@ -12,18 +12,59 @@ Copy `config.js.template` to `config.js`, and then fill in the variables. Finall
...
+`key`: key reference of this wallet in the object of created wallets. Used to say which wallet each request is directed to.
+`passphrase`: Optional parameter to start the wallet with a passphrase. If not sent we use empty string.
+`seedKey`: Optional parameter to define which seed (from the object seeds in the config file) will be used to generate the wallet. If not sent we use 'default'.
+
+```bash
+$ curl -X POST --data "key=key"  --data "passphrase=123" --data "seedKey=default" http://localhost:8000/start
+{"success":true}
+```
+
+**All requests below must have a parameter 'key' in the body which wallet must be used.**
```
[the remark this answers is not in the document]
pedroferreira1: - About having an easy way to use a single wallet:
  Yes, I was thinking about opening a new PR to handle this after we finish this one. I didn't want to make this extensive, since we need it merged by tomorrow morning.
  We could discuss this solution but I think we should have and explicit parameter in the config file to set "multiple": true/false, I think. Then, if it's a single wallet, we already get from default and start it with no need to call `/start`. But we can discuss later, I feel we should do it in a separate PR.

- About the 'key' parameter:
  I don't like having it in the header, feels like wrong to me having a parameter of the request in the header. And the decision of having it as a parameter was exactly to make it easier to handle single/multiple wallets in the same endpoint with no need to create one for each (as you described).

  If you don't like having the GET parameter in the body, I can add it to the query params, since is more common to have. It would be `http://localhost:8000/wallet/status/?key=key`. The reason I decided to use the body was just to handle the parameter in a GET and POST requests the same way. But if you feel having a body parameter for GET is not good, I can do a query param for GET and a body param for POST and handle them separately.

README.md
obiyankenobi: Yeah, it's not mentioned anywhere else. Also, it seems like an implementation detail. Not sure if it should be mentioned.

README.md
```diff
@@ -12,18 +12,59 @@ Copy `config.js.template` to `config.js`, and then fill in the variables. Finall
...
 
 Check out the full documentation in the OpenAPI Documentation in `api-docs.js`.
 
+### Start a wallet
+
+Create and start a new wallet.
+
+Parameters:
+
+`key`: key reference of this wallet in the object of created wallets. Used to say which wallet each request is directed to.
```
msbrogli: I'd say: a unique identifier for the new wallet. It will be required to access this wallet later.

README.md
msbrogli: Shouldn't we allow a `seed` parameter? In this case, the API must receive either the seed or the seedKey.

index.js
```diff
@@ -44,49 +34,126 @@ app.get('/docs', (req, res) => {
...
+    return;
+  }
+
+  // The user must send a key to index this wallet
+  if (!('key' in req.body)) {
+    res.send({
+      success: false,
+      message: 'Parameter \'key\' is required.',
+    });
+    return;
```
obiyankenobi: If we go this way of having the key in the body, I *think* there's a way to use an express middleware to handle it, so all paths that need the key go through that first.

But as we want this done by tomorrow, I'm ok if we merge like this and refactor later.

README.md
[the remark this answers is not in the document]
msbrogli: I agree with @obiyankenobi. We could get the key from the query string.

```
curl -X GET http://localhost:8000/wallet/status/?key={key}
```

index.js
```diff
@@ -44,49 +34,126 @@ app.get('/docs', (req, res) => {
...
+  }
+
+  // Passphrase is optional but if not passed as parameter
+  // the wallet will use empty string
+  if (req.body.passphrase) {
+    walletConfig['passphrase'] = req.body.passphrase;
+  }
+
+  const wallet = new HathorWallet(walletConfig);
+  wallets[req.body.key] = wallet;
```
obiyankenobi: Should it only be added after `wallet.start` succeeds, inside the promise?

README.md
msbrogli: I think it is the "set of wallets available". Maybe we should say "and remove it from the available wallets.".

README.md
```diff
@@ -190,3 +231,24 @@ $ curl http://localhost:8000/0/tx-history
...
+```bash
+$ curl -X POST --data "key=key"  http://localhost:8000/wallet/stop
+{"success":true}
+```
+
+## Util scripts
+
+### Generate new words
+
+Generate a new seed (24 words)
```
msbrogli: I'd say: "This script generates a new seed with 24 words. It is useful when you'd like to start a brand new wallet. The seed is a sensitive data that cannot be shared with anyone else. Remember to clean up your clipboard after copying & pasting."

README.md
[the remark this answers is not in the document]
obiyankenobi: I'm not sure it would be a problem having it on the header. I've seem http headers used a lot of times for authentication/authorization. But to be honest, I'm not super familiar with what type of information should be passed on the headers.

As this `key` seems a bit like authentication to me, I'd say it's not a problem to use it in the headers. But I'd wait for more opinions. @jansegre @msbrogli ?

If we go this way, having multiple/single wallets is trivial, as you can still use the store and, if the user does not send a key in the headers, use some default key like `_DEFAULT_`.

README.md
msbrogli: I think the `seedKey` must be mandatory. We should never use the `default` seed by default.

index.js
```diff
@@ -44,49 +34,126 @@ app.get('/docs', (req, res) => {
...
-  res.send({
-    'statusCode': wallet.state,
-    'statusMessage': humanState[wallet.state],
-    'network': wallet.network,
-    'serverUrl': wallet.server,
-    'serverInfo': wallet.serverInfo,
+app.post('/start', (req, res) => {
+  // We expect the user to send the seed he wants to use
+  // if not sent, we use the 'default' key
+  const seedKey = req.body.seedKey || 'default';
```
msbrogli: I think we should return an error if the seedKey is not explicitly given.

index.js
```diff
@@ -96,20 +163,22 @@ walletRouter.post('/simple-send-tx', (req, res) => {
...
   app.use(apiKeyAuth(config.http_api_key));
 }
-app.use(express.json());
-app.use(express.urlencoded({ extended: true }));
-app.use('/0', walletRouter);
+app.use('/wallet', walletRouter);
 
 app.listen(config.http_port, config.http_bind_address, () => {
   console.log(`Hathor Wallet listening on ${config.http_bind_address}:${config.http_port}...`);
-  wallet.start().then((info) => {
```
msbrogli: Maybe we should have a parameter in the config (or in the args from the cmdline) to auto start a wallet.

README.md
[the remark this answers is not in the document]
jansegre: I kind of like @obiyankenobi's idea of using a header. The `key` is more of a context indicator, and if it is *always* required it may be practical to use a header because several it is common for http libraries to allow adding headers to the "session" or "connection pool" object to be used on every request.

I just don't know if using (reusing) an authentication header is a good idea. Maybe something like `X-Hathor-Key`, `X-Hathor-Wallet`, `X-Wallet-Id` or `X-Hathor-Wallet-Id`. Being picky I think `key` is too generic and may be confused with `key/value`. Custom headers must start with `X-` (per the http spec if I'm not mistaken), and it may be good to start with `X-Hathor` but I don't know if we should go that far.

README.md
jansegre: > unique identifier for the new wallet

It sounds like `wallet-id` describes this better than `key` (which I think is too generic). If we have to name this parameter I'd prefer something like `wallet-id`. I like the idea of using headers, which would need a name for the parameter. Some alternatives are `wallet-index`/`wallet-idx` (which to me sound more like something that would be a small integer: `wallet-idx=0`, `wallet-idx=1`, ...)

An alternative would be arranging this so it doesn't need a name (like a positional argument). Maybe in the URL.

README.md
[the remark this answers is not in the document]
obiyankenobi: I agree about just using `key`. `X-Wallet-Id` or `X-Hathor-Wallet-Id` both work well.

README.md
jansegre: Is this absolutely necessary? I don't like that we are incentivizing this kind of API, it's really bad to have this level of sensitive information being passed around like this.

It's one thing to have a vulnerability that allows an attacker to make an unauthorized transaction. It is another to have a vulnerability that leeks a seed that allows a transaction at any point in the future to be made potentially without leaving traces of where de leek happened, and leaving no other fix but to change every seed ever generated while there was a leek.

If this is absolutely necessary in the immediate future, I'd rather not document it and possible add this behind a `/deprecated`, `/unsafe`, `/not-recommended` or something like that.

README.md
jansegre: I don't know, the less we exchange seeds the better.

README.md
obiyankenobi: @jansegre this is not an API, just an util script to be run locally.

index.js
pedroferreira1: Even if we go with other implementations it's important to use this middleware to handle the required parameters.

I am already doing it here (https://github.com/HathorNetwork/hathor-wallet-headless/blob/e830b4998b121e79a06373fefbe857ca1125b7dc/index.js#L94) in the `walletRouter`, so all `/wallet/...` requests don't need to check it. However, this `/start/` endpoint you added the comment is not inside `walletRouter` and I can't have it handled by a global middleware because I have other endpoints (`/` and `/docs`) that don't need the parameter.

If you know anyway I could do it better, I can refactor it.

index.js
pedroferreira1: Yeah, it's better

README.md
pedroferreira1: I agree with @jansegre that is not safe. The advantage of allowing this is that we would be able to start new wallets dynamically, only with the seeds already added in the config file. But I guess we should discuss it more, don't know if it's a strong requirement which would make us be more flexible with security.

README.md
pedroferreira1: Yes, it just logs to the user screen and nothing else. It's just a helper to generate a new seed, so users don't need to use other ways (wallet desktop, wallet mobile, or executing our method in the lib).

index.js
pedroferreira1: The user would have to select which `seedKey` and which `key` he wants to auto start. Even though I think it's good to have it, I think it's more important when you want a single wallet.

In my opinion we should discuss adding an easier support for single wallets in another PR.

README.md
pedroferreira1: Makes sense, changed it.

index.js
pedroferreira1: Makes sense, changed it.

index.js
obiyankenobi: You're right, it was my bad. I confused this other piece of code to another endpoint, so I thought we were repeating it. 

The question is why this api is not under `walletRouter`? Why don't we use `/wallet/start`?

README.md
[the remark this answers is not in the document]
jansegre: I don't think accessing it from the browser is particularly useful, assuming the only return format is JSON. If it's for development I'm pretty sure it is possible to at least resend requests with edited headers from the development tools.

But I also don't think it's a bad idea to look for multiple places, but for the sake of simpler start I'd prefer to begin with only one way to do it and add other ways in the future (specially if there's feedback that changes our view on this).

README.md
jansegre: Oh, my bad. OK then.

index.js
pedroferreira1: I defined the API's under `walletRouter` to be the ones where you already have a wallet created and want to execute an action on it. It was just the way I thought at first, the `/start/` would be to create but I can move to `/wallet/start/`

README.md
pedroferreira1: Done!

README.md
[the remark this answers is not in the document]
pedroferreira1: I agree about looking for multiple places.

My opinion is to keep only in the request for now (in case of a GET I will change to the query and for POST I will keep in the body), since it's already done and after we can add support for the header also.

What do you think?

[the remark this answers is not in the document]
obiyankenobi: I agree with Jan, I don't see why it's useful to access it in the browser at the moment. I think we should design for the use case we have right now and we can extend it in the future if there are other requirements. So the header solution is still the best, in my opinion.

index.js
obiyankenobi: I think it makes sense semantically (to have it under `/wallet/`). So we have

```
/wallet/start
/wallet/stop
/wallet/status
```

pedroferreira1: Actually to me semantically makes more sense to have under `/wallet/` some actions to be done in a wallet that you already have. And the code becomes less complex, since I can create the middleware that I already have https://github.com/HathorNetwork/hathor-wallet-headless/blob/e830b4998b121e79a06373fefbe857ca1125b7dc/index.js#L94

The piece of code below

```
  if (!(key in wallets)) {
    sendError('Invalid key parameter.')
    return;
  }

  const wallet = wallets[key];
  if (!wallet.isReady()) {
    sendError('Wallet is not ready.', wallet.state)
    return;
  }

  // Adding to req parameter, so we don't need to get it in all requests
  req.wallet = wallet;
```

is executed to all requests under `/wallet/`. I would need to add an extra check to execute this only if it's not a `/start/` because the context of `/wallet/start` is totally different from the other requests under `/wallet/`. That's why I left `/start/` outside `/wallet/`.

README.md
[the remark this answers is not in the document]
pedroferreira1: I just think that if we will add support to look for the wallet-id in multiple places, there is no reason to change what it's already coded.

Besides that, I still don't think having it in the header is better than having in a request parameter. I feel it's pretty much the same solution and, since we are thinking about the real use case, people are usually more used to send parameters than creating headers, thinking about a common use. I just don't see why it's better to have it on the header, rather than a request parameter.

index.js
obiyankenobi: Got it. I'm ok with it them.

README.md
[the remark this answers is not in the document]
obiyankenobi: I agree, we should do it only one way right now. I said we should extend only if it's required in the future.

The advantage I see in using it in the header is that it'll be handled the same way in all requests, either GET or POST. I think the more "correct" way would be having it on the path itself (`/wallet/{id}/start`, `/wallet/{id}/simple-send-tx`, etc), but that would make a bit more complex for handling both multiple and single wallets. That's why I prefer the header solution.

As for keeping this way as it's already coded, while I'm usually ok with that, this will be used externally, so it'll be harder to change moving forward. That's why I think we should get it right before releasing, given that it also wouldn't be much work.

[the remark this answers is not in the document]
pedroferreira1: I agree that it's harder to change but one of the ideas is to support multiple ways. That's why I think there is no problem if we release like that, it will continue being supported in the future.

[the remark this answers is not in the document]
pedroferreira1: Added as a header. `X-Wallet-Id`

PR #25, opened 2020-10-28
feat: add POST to create token and token as optional parameters on balance and simple-send-tx

pedroferreira1:
There are some features/fixes on this PR:

1. If I receive a `/start` to a key I already have, I stop the old wallet and start a new one. This fixes a bug when sending `/start` many times to the same key/wallet
2. Add optional token parameter  to `/balance`, `/simple-send/tx` and `/send-tx` endpoints
3. Add endpoint to create, mint and melt tokens (`/create-token`, `/mint-tokens`, `/melt-tokens`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,7 +86,7 @@ app.post('/start', (req, res) => {
     // We already have a wallet for this key
     // so we log that it won't start a new one because
     // it must first stop the old wallet and then start the new
-    console.log('Error starting wallet because this wallet-id is already used. You must stop the wallet using it first.');
+    console.log('Error starting wallet because this wallet-id is already in use. You must stop the wallet first.');
     res.send({
       success: false,
       message: `Failed to start wallet with wallet id ${walletID}`,
@@ -219,6 +219,7 @@ walletRouter.post('/simple-send-tx', (req, res) => {
 walletRouter.post('/send-tx', (req, res) => {
   const wallet = req.wallet;
   const outputs = req.body.outputs;
+  // Expects array of objects with {'hash', 'index'}
   const inputs = req.body.inputs || [];
   // Expects object with {'uid', 'name', 'symbol'}
   const token = req.body.token || null;
--- a/index.js
+++ b/index.js
@@ -34,6 +34,10 @@ app.get('/docs', (req, res) => {
   res.send(apiDocs);
 });
 
+/**
+ * POST request to start a new wallet
+ * For the docs, see api-docs.js
+ */
 app.post('/start', (req, res) => {
   // We expect the user to send the seed he wants to use
   if (!('seedKey' in req.body)) {
@@ -145,6 +149,10 @@ walletRouter.use((req, res, next) => {
   next();
 });
 
+/**
+ * GET request to get the status of a wallet
+ * For the docs, see api-docs.js
+ */
 walletRouter.get('/status', (req, res) => {
   const wallet = req.wallet;
   res.send({
@@ -156,6 +164,10 @@ walletRouter.get('/status', (req, res) => {
   });
 });
 
+/**
+ * GET request to get the balance of a wallet
+ * For the docs, see api-docs.js
+ */
 walletRouter.get('/balance', (req, res) => {
   const wallet = req.wallet;
   // Expects token uid
@@ -164,6 +176,10 @@ walletRouter.get('/balance', (req, res) => {
   res.send(balance);
 });
 
+/**
+ * GET request to get an address of a wallet
+ * For the docs, see api-docs.js
+ */
 walletRouter.get('/address', (req, res) => {
   const wallet = req.wallet;
   const index = req.query.index || null;
@@ -177,6 +193,10 @@ walletRouter.get('/address', (req, res) => {
   res.send({ address });
 });
 
+/**
+ * GET request to get all addresses of a wallet
+ * For the docs, see api-docs.js
+ */
 walletRouter.get('/addresses', (req, res) => {
   const wallet = req.wallet;
   // TODO Add pagination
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -80,6 +80,16 @@ app.post('/start', (req, res) => {
...
+  const walletID = req.body['wallet-id'];
+
+  if (walletID in wallets) {
+    // We already have a wallet for this key
+    // so we stop the old wallet and start the new one
+    const oldWallet = wallets[walletID];
+    oldWallet.stop();
+
+    delete wallets[walletID];
+  }
```
obiyankenobi, line 92: Why do it do it like this? Can't we just return saying this wallet is already started?

index.js
```diff
@@ -124,6 +134,8 @@ walletRouter.use((req, res, next) => {
     return;
   }
 
+  console.log('Received request to', req.originalUrl)
```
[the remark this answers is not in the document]
pedroferreira1: The idea here was to leave this log so we can see the requests when running the API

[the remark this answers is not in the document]
obiyankenobi: Got it, fine by me.

index.js
obiyankenobi: Ah, didn't think about this other use. Actually, I think we should then return an error if he tries to start a wallet with an id already in use. He should then explicitly stop it and start another one.

pedroferreira1: Yeah, I agree with you. Changed the code to have this behaviour.

index.js
```diff
@@ -80,6 +80,19 @@ app.post('/start', (req, res) => {
   if (req.body.passphrase) {
     walletConfig['passphrase'] = req.body.passphrase;
   }
+  const walletID = req.body['wallet-id'];
+
+  if (walletID in wallets) {
+    // We already have a wallet for this key
+    // so we log that it won't start a new one because
+    // it must first stop the old wallet and then start the new
+    console.log('Error starting wallet because this wallet-id is already used. You must stop the wallet using it first.');
```
msbrogli, line 89: Typo: `... wallet-id is already in use. You must stop the wallet first.`

index.js
```diff
@@ -185,7 +202,9 @@ walletRouter.post('/simple-send-tx', (req, res) => {
   const wallet = req.wallet;
   const address = req.body.address;
   const value = parseInt(req.body.value);
-  const ret = wallet.sendTransaction(address, value);
+  // Expects object with {'uid', 'name', 'symbol'}
+  const token = req.body.token || null;
```
msbrogli, line 206: Use `tokenId` instead of `token`?

index.js
```diff
@@ -201,7 +220,60 @@ walletRouter.post('/send-tx', (req, res) => {
   const wallet = req.wallet;
   const outputs = req.body.outputs;
   const inputs = req.body.inputs || [];
-  const ret = wallet.sendManyOutputsTransaction(outputs, inputs)
+  // Expects object with {'uid', 'name', 'symbol'}
+  const token = req.body.token || null;
```
msbrogli, line 224: Same here. `tokenId` instead of `token`.

index.js
```diff
@@ -201,7 +220,60 @@ walletRouter.post('/send-tx', (req, res) => {
...
+      res.send(response);
+    }, (error) => {
+      res.send({success: false, error});
+    });
+  } else {
+    res.send({success: false, error: ret.message});
+  }
+});
+
+walletRouter.post('/create-token', (req, res) => {
```
msbrogli, line 237: Missing docstring.

index.js
```diff
@@ -201,7 +220,60 @@ walletRouter.post('/send-tx', (req, res) => {
...
+      res.send(response);
+    }, (error) => {
+      res.send({success: false, error});
+    });
+  } else {
+    res.send({success: false, error: ret.message});
+  }
+});
+
+walletRouter.post('/mint-tokens', (req, res) => {
```
msbrogli, line 255: Missing docstring.

index.js
```diff
@@ -201,7 +220,60 @@ walletRouter.post('/send-tx', (req, res) => {
...
+      res.send(response);
+    }, (error) => {
+      res.send({success: false, error});
+    });
+  } else {
+    res.send({success: false, error: ret.message});
+  }
+});
+
+walletRouter.post('/melt-tokens', (req, res) => {
```
msbrogli, line 272: Missing docstring.

msbrogli (approved):

index.js
pedroferreira1: I agree 100% with you, however we should have a lib refactor to support this. 

I had this discussion with @obiyankenobi here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/111#discussion_r544531482

index.js
pedroferreira1: Fixed, thanks

index.js
pedroferreira1: We don't really have any default docstring for APIs, should we start adding docstring to each endpoint?
What we have here is the `api-docs.js` which was not updated and I've just added all new endpoints and updated old ones. What do you think? Should we have the api-docs file and the docstring? Or a simple docstring version?

index.js
msbrogli: Ok. Do we have another issue to refactor it later? A simple way is to create a new method that receives only the `tokenId` and make this method call the new one passing `token.uid`. What do you think?

index.js
msbrogli: You're right. I think the `api-docs.js` is enough. Maybe we can add a docstring like "For the docs, see api-docs.js". What do you think?

msbrogli (approved):

index.js
pedroferreira1: Created the issue. When we create the new method we can do exactly what you said.

index.js
pedroferreira1: Resolved in the other comment

index.js
pedroferreira1: I've added a short description of each endpoint and the reference to the docs

andreabadesso (approved):

PR #34, opened 2020-12-23
feat: add index as query parameter in address GET API

pedroferreira1:


index.js
```diff
@@ -149,8 +149,14 @@ walletRouter.get('/balance', (req, res) => {
 
 walletRouter.get('/address', (req, res) => {
   const wallet = req.wallet;
-  const markAsUsed = req.query.mark_as_used || false;
-  const address = wallet.getCurrentAddress({markAsUsed});
+  const index = req.query.index || null;
+  let address;
+  if (index) {
```
jansegre, line 154: Is `0` a valid index? Should we test for null instead implicit bool, which would put `index=0` with the same behavior as no index?

jansegre (approved):

index.js
pedroferreira1: Makes sense! Changed to `if (index !== null)`, much better

jansegre (approved):

obiyankenobi (approved):

PR #48, opened 2021-02-18
feat: utxo consolidation

thr0wn:
- New endpoint GET `/wallet/utxo-filter`
Return utxos and some helpful information regarding it.

- New endpoint POST `/wallet/utxo-consolidation`
Consolidates utxos to a given address.

- Update `api-docs`

### DIFF
--- a/api-docs.js
+++ b/api-docs.js
@@ -947,7 +947,7 @@ const apiDoc = {
           {
             name: 'max_utxos',
             'in': 'query',
-            description: 'Maximum number of utxos to aggregate. Default to MAX_INPUTS (255)',
+            description: 'Maximum number of utxos to return. Default to MAX_INPUTS (255)',
             required: false,
             schema: {
               type: 'integer',
@@ -974,7 +974,7 @@ const apiDoc = {
           {
             name: 'amount_smaller_than',
             'in': 'query',
-            description: 'Maximum limit of utxo amount to filter the utxos list. We will consolidate only utxos that have an amount lower than this value. Integer representation of decimals, i.e. 100 = 1.00.',
+            description: 'Maximum limit of utxo amount to filter the utxos list. We will return only utxos that have an amount lower than this value. Integer representation of decimals, i.e. 100 = 1.00.',
             required: false,
             schema: {
               type: 'integer',
@@ -983,7 +983,7 @@ const apiDoc = {
           {
             name: 'amount_bigger_than',
             'in': 'query',
-            description: 'Minimum limit of utxo amount to filter the utxos list. We will consolidate only utxos that have an amount bigger than this value. Integer representation of decimals, i.e. 100 = 1.00.',
+            description: 'Minimum limit of utxo amount to filter the utxos list. We will return only utxos that have an amount bigger than this value. Integer representation of decimals, i.e. 100 = 1.00.',
             required: false,
             schema: {
               type: 'integer',
@@ -992,7 +992,7 @@ const apiDoc = {
           {
             name: 'maximum_amount',
             'in': 'query',
-            description: 'Limit the maximum total amount to consolidate summing all utxos. Integer representation of decimals, i.e. 100 = 1.00.',
+            description: 'Limit the maximum total amount to return summing all utxos. Integer representation of decimals, i.e. 100 = 1.00.',
             required: false,
             schema: {
               type: 'integer',
--- a/api-docs.js
+++ b/api-docs.js
@@ -897,6 +897,210 @@ const apiDoc = {
         },
       },
     },
+    '/wallet/utxo-details': {
+      get: {
+        summary: 'Return utxos and some helpful information regarding it.',
+        parameters: [
+          {
+            name: 'x-wallet-id',
+            'in': 'header',
+            description: 'Define the key of the corresponding wallet it will be executed the request.',
+            required: true,
+            schema: {
+              type: 'string',
+            },
+          },
+          {
+            name: 'max_utxos',
+            'in': 'query',
+            description: 'Maximum number of utxos to aggregate. Default to MAX_INPUTS (255)',
+            required: false,
+            schema: {
+              type: 'integer',
+            },
+          },
+          {
+            name: 'token',
+            'in': 'query',
+            description: 'Token to filter the utxos. If not sent, we select only HTR utxos.',
+            required: false,
+            schema: {
+              type: 'string',
+            },
+          },
+          {
+            name: 'filter_address',
+            'in': 'query',
+            description: 'Address to filter the utxos.',
+            required: false,
+            schema: {
+              type: 'string',
+            },
+          },
+          {
+            name: 'amount_smaller_than',
+            'in': 'query',
+            description: 'Maximum limit of utxo amount to filter the utxos list. We will consolidate only utxos that have an amount lower than this value. Integer representation of decimals, i.e. 100 = 1.00.',
+            required: false,
+            schema: {
+              type: 'integer',
+            },
+          },
+          {
+            name: 'amount_bigger_than',
+            'in': 'query',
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -330,6 +330,36 @@ walletRouter.post('/melt-tokens', (req, res) => {
   }
 });
 
+/**
+ * GET request to get utxo details
+ * For the docs, see api-docs.js
+ */
+walletRouter.get('/utxo-details', (req, res) => {
+  const wallet = req.wallet;
+  const options = req.query || {};
```
pedroferreira1, line 339: I think we should have each query parameter separated, not expect a full object with them. The full object should be expected in the lib method only, in my opinion.

index.js
```diff
@@ -330,6 +330,36 @@ walletRouter.post('/melt-tokens', (req, res) => {
...
+  res.send(utxoDetails);
+});
+
+/**
+ * POST request to consolidate utxos
+ * For the docs, see api-docs.js
+ */
+walletRouter.post('/utxo-consolidation', async (req, res) => {
+  const wallet = req.wallet;
+  const options = req.body || {};
```
pedroferreira1, line 350: Same comment as in the utxo-details about the parameter

index.js
```diff
@@ -330,6 +330,36 @@ walletRouter.post('/melt-tokens', (req, res) => {
...
 });
 
+/**
+ * GET request to get utxo details
+ * For the docs, see api-docs.js
+ */
+walletRouter.get('/utxo-details', (req, res) => {
+  const wallet = req.wallet;
+  const options = req.query || {};
+  const utxoDetails = wallet.getUtxos(options);
```
msbrogli, line 340: Shouldn't we paginate this response? I feel a larger wallet can get hundreds of thousands of utxos and we don't want to return all of them.

index.js
```diff
@@ -330,6 +330,36 @@ walletRouter.post('/melt-tokens', (req, res) => {
   }
 });
 
+/**
+ * GET request to get utxo details
+ * For the docs, see api-docs.js
+ */
+walletRouter.get('/utxo-details', (req, res) => {
```
msbrogli, line 337: I'd call it just `GET /utxo-filter`. Isn't it more intuitive?

index.js
```diff
@@ -330,6 +330,36 @@ walletRouter.post('/melt-tokens', (req, res) => {
...
+});
+
+/**
+ * POST request to consolidate utxos
+ * For the docs, see api-docs.js
+ */
+walletRouter.post('/utxo-consolidation', async (req, res) => {
+  try {
+    const wallet = req.wallet;
+    const { destination_address, ...options } = req.body || {};
```
msbrogli, line 351: I feel it is risky to just pass the `options` to the `consolidateUtxos()` method. It is not a problem today but it can be a problem in future versions of the lib. Let's say that, in the future, we add a new option to the `consolidateUtxos()` method, and this option shouldn't be exposed to the headless wallet. In this case, it would be exposed and we probably wouldn't notice until a bug is reported.

In other words, I feel we should only pass the valid options to the `consolidateUtxos()` methods. What do you think?

msbrogli (changes requested):

index.js
thr0wn: You are right. To allow only valid options I added `express-validator` to the middlewares chain, it will return validation errors to invalid or missing `body parameters` too (see it [here](https://github.com/HathorNetwork/hathor-wallet-headless/pull/48/commits/d0a2f74c283b31ca6d01e3f1ac9a220bd78ac407#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R355)). To me it seems to solve two problems at once, query/body properties without validation and unknown property being passed to `consolidateUtxos`. Do you like the idea?

index.js
thr0wn: Done

index.js
thr0wn: I added a `TODO` regarding this scenario. Do you think that is enough for now?

pedroferreira1: I think a `TODO` is enough for now. It's possible that we have a wallet with hundreds of thousands of utxos but really unlikely. After the wallet-service integration it can be implemented.

api-docs.js
```diff
@@ -897,6 +897,210 @@ const apiDoc = {
...
+            description: 'Define the key of the corresponding wallet it will be executed the request.',
+            required: true,
+            schema: {
+              type: 'string',
+            },
+          },
+          {
+            name: 'max_utxos',
+            'in': 'query',
+            description: 'Maximum number of utxos to aggregate. Default to MAX_INPUTS (255)',
```
pedroferreira1, line 916: `Not exactly 'to aggregate' in the GET. Can be only 'Maximum number of utxos to return. Default to MAX_INPUTS (255)'. What do you think?

api-docs.js
```diff
@@ -897,6 +897,210 @@ const apiDoc = {
...
+            description: 'Address to filter the utxos.',
+            required: false,
+            schema: {
+              type: 'string',
+            },
+          },
+          {
+            name: 'amount_smaller_than',
+            'in': 'query',
+            description: 'Maximum limit of utxo amount to filter the utxos list. We will consolidate only utxos that have an amount lower than this value. Integer representation of decimals, i.e. 100 = 1.00.',
```
pedroferreira1, line 943: 'We will consolidate ...' -> 'We will return ...'

api-docs.js
```diff
@@ -897,6 +897,210 @@ const apiDoc = {
...
+            required: false,
+            schema: {
+              type: 'integer',
+            },
+          },
+          {
+            name: 'amount_bigger_than',
+            'in': 'query',
+            description: 'Minimum limit of utxo amount to filter the utxos list. We will consolidate only utxos that have an amount bigger than this value. Integer representation of decimals, i.e. 100 = 1.00.',
```
pedroferreira1, line 952: 'We will consolidate ...' -> 'We will return ...'

api-docs.js
```diff
@@ -897,6 +897,210 @@ const apiDoc = {
...
+            required: false,
+            schema: {
+              type: 'integer',
+            },
+          },
+          {
+            name: 'maximum_amount',
+            'in': 'query',
+            description: 'Limit the maximum total amount to consolidate summing all utxos. Integer representation of decimals, i.e. 100 = 1.00.',
```
pedroferreira1, line 961: 'to consolidate ...' -> 'to return ...'

pedroferreira1 (approved):

api-docs.js
thr0wn: Done, thanks!

api-docs.js
thr0wn: Done, thanks!

api-docs.js
thr0wn: Done, thanks!

api-docs.js
thr0wn: I think you are absolutely right, thanks!

index.js
msbrogli: Seems good to me. Thanks!

index.js
msbrogli: I feel we should put a limit and flag if the limit has been reached. I don't feel it is reasonable to return 1 million utxos in any case. Maybe a limit of 10,000 utxos? And the `TODO` as well, of course.

thr0wn: @msbrogli, after reviewing the filtering behavior, I noted that `max_utxos` may already serve as limiting because it defines the number of returned `utxos` and it has a default value of 255. So, even if the request does not specify `max_utxos`, it will return only 255 `utxos`. Even with `max_utxos`, should I add a new limit flag?

pedroferreira1 (approved):

index.js
msbrogli: You're right. It will return at most 255 `utxos` unless the user increases the `max_utxos` argument. Well, as it is already protected, I feel we can add pagination later.

msbrogli (approved):

PR #57, opened 2021-02-28
feat: add api to get a transaction from the wallet

pedroferreira1:


index.js
```diff
@@ -225,6 +225,21 @@ walletRouter.get('/tx-history', (req, res) => {
   }
 });
 
+/**
+ * GET request to get a transaction from the wallet
+ * For the docs, see api-docs.js
+ */
+walletRouter.get('/transaction', (req, res) => {
```
msbrogli, line 232: Missing sanitization.

index.js
```diff
@@ -225,6 +225,21 @@ walletRouter.get('/tx-history', (req, res) => {
   }
 });
 
+/**
+ * GET request to get a transaction from the wallet
+ * For the docs, see api-docs.js
```
msbrogli, line 230: Missing update in the api-docs.js.

msbrogli (changes requested):

index.js
pedroferreira1: Done

index.js
pedroferreira1: Done!

msbrogli (approved):

PR #61, opened 2021-03-02
feat: protect send-tx and simple-send-tx from being called concurrently

pedroferreira1:


index.js
```diff
@@ -16,6 +16,54 @@ import apiKeyAuth from './api-key-auth';
 import logger from './logger';
 import version from './version';
 
+// Variable that saves when a tx is being sent
+// We don't support sending multiples transactions concurrently
+// so we use this to prevent a user from sending multiples requests
+// while the first was not finished
+let sendingTx = false;
```
msbrogli, line 23: The implementation seems good to me. Just a few comments:

1) Rename variables and methods to a lock system. So, `sendingTx` would become `lock_status`, and `startSendTx` would become `getLock({timeout=SEND_TX_LOCK_TIMEOUT})`.

2) Add a comment that the lock is local and only works with one thread and no balance between multiple instances. In those cases, we would need to add support for a distributed lock (e.g. using redis).

msbrogli (changes requested):

src/lock.js
```diff
@@ -0,0 +1,66 @@
...
+    return true;
+  }
+}
+
+// Lock status enum
+// SENDING_TX is used to lock when a tx is being sent
+// We don't support sending multiples transactions concurrently
+// so we use this to prevent a user from sending multiples requests
+// while the first was not finished
+Lock.SENDING_TX = 0
```
andreabadesso, line 62: Why is this outside of the class?

pedroferreira1: Actually I had to create a const variable outside the class because it's now a singleton. Can you review again please?

msbrogli (approved):

index.js
msbrogli: Done! Thanks!

andreabadesso (approved):

PR #62, opened 2021-03-02
feat: all integer parameters should be sanitized toInt before usage

pedroferreira1:


index.js
```diff
@@ -381,7 +382,8 @@ walletRouter.post('/send-tx',
         options: {
           min: 1
         }
-      }
+      },
+      toInt: true
     },
     inputs: {
       in: ['body'],
```
msbrogli, line 389: Don't we have to convert the `input[].index` to int as well?

index.js
```diff
@@ -202,7 +202,7 @@ walletRouter.get(
  * For the docs, see api-docs.js
  */
 walletRouter.get('/address',
-  query('index').isInt().optional(),
+  query('index').isInt().optional().toInt(),
   query('mark_as_used').isBoolean().optional(),
```
msbrogli, line 206: `toBoolean()`? https://github.com/validatorjs/validator.js#sanitizers

index.js
```diff
@@ -582,12 +584,12 @@ walletRouter.post('/melt-tokens',
...
+  query('max_utxos').isInt().optional().toInt(),
   query('token').isString().optional(),
   query('filter_address').isString().optional(),
-  query('amount_smaller_than').isInt().optional(),
-  query('amount_bigger_than').isInt().optional(),
-  query('maximum_amount').isInt().optional(),
+  query('amount_smaller_than').isInt().optional().toInt(),
+  query('amount_bigger_than').isInt().optional().toInt(),
+  query('maximum_amount').isInt().optional().toInt(),
   query('only_available').isBoolean().optional(),
```
msbrogli, line 593: `toBoolean()`? https://github.com/validatorjs/validator.js#sanitizers

msbrogli (changes requested):

index.js
pedroferreira1: Makes sense. Because of the way I had to validate the inputs, the conversion was a bit different.

index.js
pedroferreira1: Done

index.js
pedroferreira1: Done

msbrogli (approved):

index.js
msbrogli: Ok. Great!

thr0wn (approved):

PR #65, opened 2021-03-04
chore: Add Docker

### LINKED ISSUE #64: Dockerfile
Create Dockerfile, create docker-compose.yml example, and upload image to Docker Hub.

msbrogli:
Fixes #64


pedroferreira1 (approved):

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM $IMAGE_REF as stage-0
+WORKDIR /usr/src/app/
+COPY package.json package-lock.json ./
+RUN npm install
+
+# final stage
+FROM $IMAGE_REF
+WORKDIR /usr/src/app/
+COPY --from=stage-0 /usr/src/app/node_modules/ ./node_modules/
+COPY .babelrc api-docs.js api-key-auth.js index.js logger.js version.js package.json package-lock.json utils/ ./
```
jansegre, line 13: Do we need all of those (maybe `.babelrc` is the one that looks like isn't doing much to me)?

I just noticed we have a lot of loose source files, we could use a refactor to put them on a `src` dir or something.

Also I think it doesn't look like `stage-0` is saving any space. I think a single stage would be fine since we don't have any build dependencies.

Makefile
```diff
@@ -0,0 +1,23 @@
+
+.PHONY: all
+all:
+	@echo "No default action."
+
+# docker:
+
+docker_subtag := $(shell git describe --tags --dirty)
```
jansegre, line 8: I'd prefix this with `dev-` just to make sure it won't replace production build is we run `make docker-push` on a git tag commit, we can still explicitly do this with `make docker-push docker_subtag=vX.Y.Z` but in any case I'd prefer having DockerHub autobuild this or at most a Github Action like we have on `hathor-core`.

My experience with this is that we can have outdated cached layers that could break the build for subtle reasons (this happened a lot with hathor-core). So it's always best to let CI make the production builds, also makes it easier to make multi-arch builds but that's not relevant right now.

jansegre (approved):
I made a few small comments but I really liked the  `config.js.docker` approach.

Dockerfile
msbrogli: We need `.babelrc` so `npm start` runs properly. I just did the refactor and copied the `src/` dir. I also removed the stage-0. Can you review again, please?

jansegre (approved):

Dockerfile
jansegre: Got it. Looks great!

Makefile
msbrogli: Done! Thanks. We still need to configure DockerHub Autobuild or Github Action to do it.

PR #69, opened 2021-03-09
feat: Integration testing

thr0wn:
- [x] Increase the coverage
- [x] Sanitize package-lock.json
- [x] Solve conflicts


pedroferreira1:
See this comment here: https://github.com/HathorNetwork/hathor-wallet-headless/pull/79#discussion_r592930918
We must have a test for the situation in this discussion.

thr0wn:
> See this comment here: [#79 (comment)](https://github.com/HathorNetwork/hathor-wallet-headless/pull/79#discussion_r592930918)
> We must have a test for the situation in this discussion.

It is done here. Does this test seem fine to you?

pedroferreira1:
> It is done [here](https://github.com/HathorNetwork/hathor-wallet-headless/pull/69/files#diff-957965e9c38f0fdccef8502ea40196a2b4bb10f046a560a626fb9e383765571aR12). Does this test seem fine to you?

It's not enough because the comment I referenced is about a PR that the `/wallet/address?index` was failing only when index was 0. If you haven't rebased your branch yet with `dev` your `/wallet/address?index=0` will give you a wrong address. It's important to add the test also for index=0, see that it fails and then rebase your branch to see the test succeed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm install
+      - name: Test
+        run: npm run test
```
msbrogli, line 29: What about the coverage report? It would be great to upload it somewhere.

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
...
+      matrix:
+        node-version: [8.x, 10.x, 12.x, 14.x]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm install
```
msbrogli, line 27: I feel we should run a linter as well if we have any.

__tests__/address.test.js
```diff
@@ -0,0 +1,48 @@
+import TestUtils from "./test-utils";
+
+describe("address api", () => {
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .get("/wallet/address")
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.address).toBe("WmtWgtk5GxdcDKwjNwmXXn74nQWTPWhKfx");
```
msbrogli, line 9: It should return the same address if called again, right? I think we should test it as well.

__tests__/address.test.js
```diff
@@ -0,0 +1,48 @@
...
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.body.address).toBe("WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc");
+  });
+
+  it("should return a new address with mark_as_used", async () => {
+    let response = await TestUtils.request
+      .get("/wallet/address")
+      .query({ mark_as_used: true })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
```
msbrogli, line 41: The expected address here is `WmtWgtk5GxdcDKwjNwmXXn74nQWTPWhKfx`, right? We should check it as well.

__tests__/addresses.test.js
```diff
@@ -0,0 +1,11 @@
+import TestUtils from "./test-utils";
+
+describe("addresses api", () => {
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .get("/wallet/addresses")
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.addresses.length).toBeGreaterThan(20);
```
msbrogli, line 9: I feel the addresses are ordered by their indexes, right? In this case, we should check the first 5 address just to be sure.

__tests__/address.test.js
```diff
@@ -0,0 +1,48 @@
...
+      .get("/wallet/address")
+      .query({ index: "2" })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.body.address).toBe("WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc");
+  });
+
+  it("should return a new address with mark_as_used", async () => {
+    let response = await TestUtils.request
+      .get("/wallet/address")
+      .query({ mark_as_used: true })
```
msbrogli, line 39: If we call this API enough times with `mark_as_used`, it will return the same address over and over after the gapLimit is reached. I feel we need to add a test for it.

__tests__/create-token.test.js
```diff
@@ -0,0 +1,71 @@
...
+    const response = await TestUtils.request
+      .post("/wallet/create-token")
+      .send({
+        name: "stub_token",
+        symbol: "03",
+        amount: 1,
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
```
msbrogli, line 14: Should we check for success?

__tests__/create-token.test.js
```diff
@@ -0,0 +1,71 @@
...
+    const response = await TestUtils.request
+      .post("/wallet/create-token")
+      .send({
+        name: "stub_token",
+        symbol: "03",
+        amount: "1",
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
```
msbrogli, line 27: Shouldn't we check for success?

__tests__/create-token.test.js
```diff
@@ -0,0 +1,71 @@
...
+    const promise2 = TestUtils.request
+      .post("/wallet/create-token")
+      .send({
+        name: "stub_token",
+        symbol: "03",
+        amount: 1,
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+
+    const [response1, response2] = await Promise.all([promise1, promise2]);
```
msbrogli, line 65: I feel we should add tests for a high amount such as there's no HTR funds to do it, and also to try to create a token with the symbol HTR or the name Hathor.

__tests__/melt-tokens.test.js
```diff
@@ -0,0 +1,67 @@
...
+      const response = await TestUtils.request
+        .post("/wallet/melt-tokens")
+        .send(token)
+        .set({ "x-wallet-id": TestUtils.walletId });
+      expect(response.status).toBe(400);
+      expect(response.body.success).toBeFalsy();
+    });
+  });
+
+  it("should receive an error when trying to do concurrent melt-tokens (lock/unlock behavior)", async () => {
```
msbrogli, line 45: I feel we should test melting less than 1 token, so there won't be any HTR in return of the melt operation. It was a bug in previous versions of the wallet.

__tests__/mint-tokens.test.js
```diff
@@ -0,0 +1,66 @@
...
+      .post("/wallet/mint-tokens")
+      .send({
+        token: "03",
+        amount: "1",
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
+  });
+
```
msbrogli, line 27: I feel we should test minting 0.01 tokens, which will require a 0.01 HTR deposit.

__tests__/send-tx.test.js
```diff
@@ -0,0 +1,121 @@
...
+      .post("/wallet/send-tx")
+      .send({
+        outputs: [
+          { address: "WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc", value: 1 },
+          { address: "WewDeXWyvHP7jJTs7tjLoQfoB72LLxJQqN", value: "1" },
+        ],
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
```
msbrogli, line 26: I feel we should check the success of the operation, shouldn't we?

__tests__/simple-send-tx.test.js
```diff
@@ -0,0 +1,125 @@
...
+          name: "stub_token_2",
+          uid: "02",
+          symbol: "stub_token_2",
+        },
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBeFalsy();
+  });
+
```
msbrogli, line 104: I feel we should test that it fails when the change address does not belong to the wallet.

__tests__/utxo.test.js
```diff
@@ -0,0 +1,26 @@
...
+    const response = await TestUtils.request
+      .post("/wallet/utxo-consolidation")
+      .send({
+        destination_address: "WWbt2ww4W45YLUAumnumZiyWrABYDzCTdN",
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.total_utxos_consolidated).toBe(13);
+    expect(response.body.total_amount).toBe(83199);
+    expect(response.body.utxos).toHaveLength(13);
```
msbrogli, line 24: I feel we should include another call to the `/wallet/utxo-filter` to double check the number of utxos has decreased. And also confirm that the wallet's balance is still the same.

src/index.js
```diff
@@ -786,6 +786,10 @@ if (config.gapLimit) {
   walletUtils.setGapLimit(config.gapLimit);
 }
 
-app.listen(config.http_port, config.http_bind_address, () => {
-  console.log(`Listening on ${config.http_bind_address}:${config.http_port}...`);
-});
+if (process.env.NODE_ENV !== 'test') {
```
msbrogli, line 789: I fell we should move the `app` creation to another file and import it here. So, the tests can easily create a new app just importing from this new file. Maybe we should do it in a separate PR.

msbrogli (changes requested):

__tests__/addresses.test.js
```diff
@@ -0,0 +1,11 @@
+import TestUtils from "./test-utils";
+
+describe("addresses api", () => {
```
pedroferreira1, line 3: We should try to create another test here where we set a new gap limit and see that the addresses loaded is greater than this value.

__tests__/balance.test.js
```diff
@@ -0,0 +1,32 @@
+import TestUtils from "./test-utils";
+
+describe("balance api", () => {
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .get("/wallet/balance")
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.available).toBe(83199);
+    expect(response.body.locked).toBe(0);
```
pedroferreira1, line 10: Maybe we should add a locked utxo in the fixtures to test this. All balances have locked as 0

__tests__/create-token.test.js
```diff
@@ -0,0 +1,71 @@
...
+      const response = await TestUtils.request
+        .post("/wallet/create-token")
+        .send(token)
+        .set({ "x-wallet-id": TestUtils.walletId });
+      expect(response.status).toBe(400);
+      expect(response.body.success).toBeFalsy();
+    });
+  });
+
+  it("should receive an error when trying to do concurrent create-token (lock/unlock behavior)", async () => {
```
pedroferreira1, line 47: Can we create a test to try to run `/wallet/create-token` and `/wallet/simple-send-tx` in the same time? It should block the second request as well.

__tests__/melt-tokens.test.js
pedroferreira1: Amount 1 is actually 0.01, so I guess it's already tested.

I think we should add a test with 0 amount, to validate it won't execute.

And add test with 100, to see the return deposit.

__tests__/melt-tokens.test.js
```diff
@@ -0,0 +1,67 @@
...
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .post("/wallet/melt-tokens")
+      .send({
+        token: "00da712d64e04866c8c9aa8fceca70e80d1693864176b6b443220cf29adab5ed",
+        amount: 1,
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
```
pedroferreira1, line 14: I think we should validate the return output has only a melt authority. Not the deposit because you've melted 0.01

__tests__/mint-tokens.test.js
```diff
@@ -0,0 +1,66 @@
...
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .post("/wallet/mint-tokens")
+      .send({
+        token: "03",
+        amount: 1,
+      })
+      .set({ "x-wallet-id": TestUtils.walletId });
+    expect(response.status).toBe(200);
+    expect(response.body.hash).toBeDefined();
```
pedroferreira1, line 13: I think you should validate you have an output with mint authority and another one with the 1 amount of the minted token.

__tests__/utxo.test.js
pedroferreira1: And double check the new utxo is in the correct destination address

package-lock.json
```diff
@@ -517,6 +517,15 @@
         "@babel/helper-plugin-utils": "^7.8.0"
```
pedroferreira1, line 517: Just a reminder to revisit this and leave only the added dependencies

__tests__/simple-send-tx.test.js
```diff
@@ -0,0 +1,125 @@
+import TestUtils from "./test-utils";
+
+describe("simple-send-tx api", () => {
+  it("should return 200 with a valid body", async () => {
+    const response = await TestUtils.request
+      .post("/wallet/simple-send-tx")
+      .send({
+        address: "WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc",
+        value: 1,
```
pedroferreira1, line 9: Can you add a test with value: 0?

__tests__/simple-send-tx.test.js
pedroferreira1: And another one with correct change address and validate the change output really went there.

__tests__/send-tx.test.js
```diff
@@ -0,0 +1,121 @@
...
+        .send({
+          outputs: [{ address: "WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc", value: 1 }],
+          token: token,
+        })
+        .set({ "x-wallet-id": TestUtils.walletId });
+      expect(response.status).toBe(400);
+      expect(response.body.success).toBeFalsy();
+    });
+  });
+});
```
pedroferreira1, line 121: I think we should add more tests here:

1. Test using a specific input (and validate it really used it)
2. Test using an invalid change address
3. Test using a valid change address (and validate that the change output went to it)
4. Test invalid value, e.g. 0

pedroferreira1 (changes requested):

__tests__/melt-tokens.test.js
msbrogli: You're right. Thanks!

__tests__/address.test.js
thr0wn: Done

__tests__/address.test.js
thr0wn: Done

__tests__/addresses.test.js
thr0wn: Done

__tests__/address.test.js
thr0wn: Done

__tests__/create-token.test.js
thr0wn: Done

__tests__/create-token.test.js
thr0wn: Done

__tests__/send-tx.test.js
thr0wn: Done

__tests__/addresses.test.js
thr0wn: Done

__tests__/create-token.test.js
thr0wn: Done

__tests__/simple-send-tx.test.js
thr0wn: Done

.github/workflows/main.yml
thr0wn: Currently, there is no linter, but we may add it on other PR.

.github/workflows/main.yml
thr0wn: Coverage report uploaded to [codecov](https://codecov.io/gh/HathorNetwork/hathor-wallet-headless/tree/744fc3a14c09fd57e56d35a44f49f37dee41c846/src).

__tests__/balance.test.js
thr0wn: Done

__tests__/create-token.test.js
thr0wn: 1. `tests for a high amount such as there's no HTR funds to do it`
Done
2. `try to create a token with the symbol HTR or the name Hathor`
It seems that this validation is done by the `fullnode`, so I can only mock this behavior. But mocking this `fullnode` error to assert it is retransmitted to `hathor-wallet-headless` does not sound like a valid effort.

__tests__/melt-tokens.test.js
thr0wn: 1. `I think we should add a test with 0 amount, to validate it won't execute.`
Done
2. `And add a test with 100, to see the return deposit.`
I`m unable to test it without mocking the `fullnode` response, but doing by it I will be only testing a mocked response.

__tests__/mint-tokens.test.js
thr0wn: I am already testing minting 0.01 tokens (amount: 1), but as the `fullnode` response is mocked, I am unable to assert the 0.01 HTR deposit, unless I mock a response exactly with 0.01 HTR deposit, but it seems to be an invalid effort.

__tests__/simple-send-tx.test.js
thr0wn: 1. `fails when the change address does not belong to the wallet`
Done
2. `succeeds with correct change address`
Done
3. `validate the change output really went there`
I can only test it mocking the fullnode response, but by doing it I will be only testing a mocked response.

__tests__/utxo.test.js
thr0wn: It requires mocking websocket messages coming from the `fullnode` that can be done later [here](https://github.com/HathorNetwork/hathor-wallet-headless/issues/88).

__tests__/melt-tokens.test.js
thr0wn: Testing it with a mocked`fullnode` is not a valid effort for now.

__tests__/mint-tokens.test.js
thr0wn: You are right, but testing it with a mocked fullnode is not a valid effort for now.

__tests__/send-tx.test.js
thr0wn: 1. Test using a specific input (and validate it really used it)
Testing it with a mocked fullnode is not a valid effort for now.
2. Test using an invalid change address
Done
3. Test using a valid change address ~~(and validate that the change output went to it)~~
Done
4. Test invalid value, e.g. 0
Done

src/index.js
thr0wn: Noted here https://github.com/HathorNetwork/hathor-wallet-headless/issues/89 to be done later.

package-lock.json
thr0wn: Some dependencies were added because the new `devDependencies` requires it as `dependencies` or as `peerDependencies`. Do you recommend removing it?

thr0wn: Done

__tests__/send-tx.test.js
```diff
@@ -0,0 +1,159 @@
...
+          outputs: [{ address: "WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc", value: 1 }],
+          token: token,
+        })
+        .set({ "x-wallet-id": TestUtils.walletId });
+      expect(response.status).toBe(400);
+      expect(response.body.success).toBeFalsy();
+    });
+  });
+
+  it("should accept a transaction with a change_address that does not belong to the wallet", async () => {
```
pedroferreira1, line 136: with a change_address that does belong 

Remove *not*

__tests__/simple-send-tx.test.js
```diff
@@ -0,0 +1,192 @@
...
+          value: 1,
+          token: token,
+        })
+        .set({ "x-wallet-id": TestUtils.walletId });
+      expect(response.status).toBe(400);
+      expect(response.body.success).toBeFalsy();
+    });
+  });
+
+  it("should accept a transaction with a change_address that does not belong to the wallet", async () => {
```
pedroferreira1, line 167: with a change_address that does belong to the wallet

Remove *not*

pedroferreira1 (approved):

msbrogli (approved):

PR #78, opened 2021-03-12
feat: create endpoint to get address index

pedroferreira1:


src/index.js
```diff
@@ -227,6 +227,27 @@ walletRouter.get('/address',
...
+  query('address').isString(),
+  (req, res) => {
+  const validationResult = parametersValidation(req);
+  if (!validationResult.success) {
+    return res.status(400).json(validationResult);
+  }
+  const wallet = req.wallet;
+  const address = req.query.address;
+  const index = wallet.getAddressIndex(address);
+  if (index) {
```
jansegre, line 244: `0` is a valid index right? But I suspect this is a string so this isn't a problem.

jansegre (approved):

src/index.js
pedroferreira1: Yes, it's a valid index and is a problem because the return of `getAddressIndex` is an integer. I've updated the code to fix this. Can you see again please?

msbrogli (approved):

src/index.js
jansegre: Great!

PR #79, opened 2021-03-12
fix: fix address generation for index 0

pedroferreira1:
We used to get `req.query.index` as string, thus when comparing `const index = req.query.index || null`, when `index = '0'` it used to work fine. However, we've added a sanitization and `toInt` protection. After that, `req.query.index` is already an integer and if `index = 0` the comparison does not work.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -216,9 +216,10 @@ walletRouter.get('/address',
     return res.status(400).json(validationResult);
   }
   const wallet = req.wallet;
-  const index = req.query.index || null;
+  const index = req.query.index;
   let address;
-  if (index !== null) {
+  if (index !== undefined) {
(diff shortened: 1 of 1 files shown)


jansegre (approved):

src/index.js
```diff
@@ -216,9 +216,10 @@ walletRouter.get('/address',
     return res.status(400).json(validationResult);
   }
   const wallet = req.wallet;
-  const index = req.query.index || null;
```
msbrogli, line 219: Nice catch! Maybe add a test to this in the other issue, right?

msbrogli (approved):

src/index.js
pedroferreira1: It's important. I will reference this comment here in the test PR

PR #82, opened 2021-03-12
feat: add allowPassphrase to docker config file

pedroferreira1:


jansegre (approved):

jansegre (approved):

config.js.docker
```diff
@@ -8,6 +8,29 @@ function getParam(name, _default) {
...
+function getBooleanParam(name, _default) {
+  const param = getParam(name, _default);
+  if (param === 'true' || param === '1') {
+    // Should we allow more representations?
+    return true;
+  } else if (param === 'false' || param === '0') {
+    // Should we allow more representations?
+    return false;
+  } else {
+    return _default;
```
msbrogli, line 30: I'd say throw an exception to prevent running with wrong parameters.

config.js.docker
```diff
@@ -8,6 +8,29 @@ function getParam(name, _default) {
   return argv[name] || process.env[envname] || _default;
 }
 
+function getIntParam(name, _default) {
+  const param = getParam(name, _default);
+  const parsedParam = parseInt(param);
+  if (Number.isInteger(parsedParam)) {
```
msbrogli, line 14: I feel `parseInt()` will always return an int. Should we prevent things like `"123a"`?

config.js.docker
```diff
@@ -8,6 +8,29 @@ function getParam(name, _default) {
...
+    return parsedParam;
+  } else {
+    return _default;
```
msbrogli, line 17: I'd say throw an exception to prevent running with wrong parameters.

msbrogli (changes requested):

config.js.docker
pedroferreira1: If this is an optional parameter we can't accept only 'true', 'false', '0' and '1'. We must accept the default parameter, which is `undefined` or `null`. That's why I kept the `else`

config.js.docker
pedroferreira1: If this is an optional parameter we must allow the default as `undefined` or `null`. Should we start with this validation in the first line?

```
if (param === undefined || param === null) return param;
```

config.js.docker
pedroferreira1: Added this commit https://github.com/HathorNetwork/hathor-wallet-headless/pull/82/commits/8e1208b58469e937e96363ab4f25dc1c15b1c3bb. What do you think?

config.js.docker
msbrogli: Good one. Thanks!

config.js.docker
msbrogli: But you also return the default value for any random string (e.g. "a").

msbrogli (approved):

config.js.docker
pedroferreira1: Yeah, I've done the same validation for the integer and boolean methods. Can you see this commit? https://github.com/HathorNetwork/hathor-wallet-headless/pull/82/commits/212631446fdf39a2d87ca54e2c499cbc9884b8c0

config.js.docker
pedroferreira1: Yeah, I've done the same validation for the integer and boolean methods. Can you see this commit? https://github.com/HathorNetwork/hathor-wallet-headless/pull/82/commits/212631446fdf39a2d87ca54e2c499cbc9884b8c0

msbrogli (approved):

PR #84, opened 2021-03-15
feat: adding GET /address-info

thr0wn:


src/index.js
```diff
@@ -747,6 +747,35 @@ walletRouter.post(
...
+/**
+ * GET request to obtain adress information
+ * For the docs, see api-docs.js
+ */
+ walletRouter.get(
+  '/address-info',
+  query('address').isString(),
+  query('token').isString().optional(),
+  async (req, res) => {
+    try {
```
msbrogli, line 759: Why do we have this try? Which exception are we expecting to handle?

src/index.js
```diff
@@ -747,6 +747,35 @@ walletRouter.post(
   }
 );
 
+/**
+ * GET request to obtain adress information
+ * For the docs, see api-docs.js
```
msbrogli, line 752: Missing update in the api-docs.js.

msbrogli (changes requested):

src/index.js
thr0wn: api-docs.js updated, thanks!

src/index.js
thr0wn: If the specified address does not belong to the wallet an error will be fired. To correctly respond to the api caller, this try catch statement catches the error and format the return.

pedroferreira1: Maybe we should capture the specific error

pedroferreira1 (approved):

src/index.js
msbrogli: Exactly. We must avoid capturing generic exceptions at all costs.

thr0wn: Done, thanks!

src/index.js
```diff
@@ -747,6 +747,39 @@ walletRouter.post(
...
+    try {
+      // Query parameters validation
+      const validationResult = parametersValidation(req);
+      if (!validationResult.success) {
+        return res.status(400).json(validationResult);
+      }
+
+      const wallet = req.wallet;
+      const { address, token } = matchedData(req, { locations: ['query'] });
+      const result = wallet.getAddressInfo(address, { token });
```
msbrogli, line 768: I feel this is the place where the exception might happen. So, we should keep the try-catch block just around it.

src/index.js
```diff
@@ -747,6 +747,39 @@ walletRouter.post(
...
+      res.send({
+        success: true,
+        ...result
+      });
+    } catch(error) {
+      if (error instanceof errors.AddressError) {
+        res.send({ success: false, error: error.message });
+      }
+
+      throw error;
```
msbrogli, line 778: Shouldn't it be in the `else` block? I mean, we shouldn't re-raise the exception if it was handled returning the error to the user, right?

msbrogli (changes requested):

src/index.js
thr0wn: Done, thanks.

src/index.js
thr0wn: Absolutely right, thanks.

msbrogli (approved):

msbrogli (approved):

PR #101, opened 2021-06-21
feat: send tx now has support for filling inputs with utxos filter

pedroferreira1:


src/index.js
```diff
@@ -415,6 +415,70 @@ walletRouter.post('/simple-send-tx',
...
+  // If I can't fill all the amount with the returned utxos, then return null
+  if (utxosDetails.total_amount_available < sumOutputs) {
+    return null;
+  }
+
+  const utxos = utxosDetails.utxos;
+  // Sort utxos with larger amounts first
+  utxos.sort((a, b) => b.amount - a.amount);
+
+  if (utxos[0].amount > sumOutputs) {
```
msbrogli, line 452: I feel we already have something similar when selecting utxos without queries. Can't we reuse that piece of code?

src/index.js
```diff
@@ -453,18 +517,24 @@ walletRouter.post('/send-tx',
       isObject: true,
       custom: {
         options: (value, { req, location, path }) => {
-          if (!('hash' in value) || !(typeof value.hash === 'string')) {
-            return false;
-          }
-          if (!('index' in value) || !(/^\d+$/.test(value.index))) {
-            // Test that index is required and it's an integer
-            return false;
+          if ('type' in value && value.type === 'query') {
```
msbrogli, line 520: Missing docs explaining how to use it and what query parameters are supported. I feel that we should include at least one example.

msbrogli (approved):

src/index.js
pedroferreira1: We have a method that does something similar in the lib. The method is `getInputsFromAmount` from `src/wallet.js`.

However this method can't be used as we want and, as I described in the PR, the best approach would be to create a new method in the lib for this but the current version of the lib is not compatible with the headless because of other refactors. The work to make it compatible is not small and might be lost (because the refactor in the facades is not over yet). That's why I decided to create this new method in the headless itself.

src/index.js
pedroferreira1: Added api docs for it with parameter options.

msbrogli: Great! Thanks!

src/index.js
msbrogli: Ok. So this code duplication will be fixed when we migrate this part to the lib, right? Have you opened an issue to get this migration done later?

msbrogli (approved):

src/index.js
pedroferreira1: Yes.

andreabadesso (approved):

PR #112, opened 2021-08-08
refactor: make wallet headless compatible with lib again

pedroferreira1:
TODO

- [x] Missing update package.json with new library version (waiting for lib release)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -126,9 +126,13 @@ app.post('/start', (req, res) => {
   }
 
   const connection = new Connection({network: config.network, servers: [config.server], connectionTimeout: config.connectionTimeout});
+  // Previous versions of the lib would have password and pin default as '123'
+  // We currently need something to be defined, otherwise we get an error when starting the wallet
   const walletConfig = {
     seed,
-    connection
+    connection,
+    password: '123',
+    pinCode: '123',
   }
 
   // tokenUid is optionat but if not passed as parameter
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -126,9 +126,13 @@ app.post('/start', (req, res) => {
   }
 
   const connection = new Connection({network: config.network, servers: [config.server], connectionTimeout: config.connectionTimeout});
+  // Previous versions of the lib would have password and pin default as '123'
+  // We currently need something to be defined, otherwise we get an error when starting the wallet
   const walletConfig = {
     seed,
-    connection
+    connection,
+    password: '123',
```
msbrogli, line 134: Have we removed the default pin and password in the newer versions of the lib? I guess we should add an option to skip storing the seed and other critical data in the storage.

msbrogli (approved):

src/index.js
pedroferreira1: Yes, we've removed the default because it's not required anymore. In the wallets, sometimes we are starting the HathorWallet object but the sensitive data is already encrypted on storage, then we don't need pin/password to create the wallet object

pedroferreira1: Yes, we've removed the default because it's not required anymore. In the wallets, sometimes we are starting the HathorWallet object but the sensitive data is already encrypted on storage, then we don't need pin/password to create the wallet object

msbrogli: Ok. For the headless wallet, these two options are not critical since the storage is just in memory.

andreabadesso (approved):

PR #118, opened 2021-08-19
refactor: adapting wallet facade calls to use new method signatures and returns

pedroferreira1:
TODO

- [x] Stil need to bump wallet lib version to work

### DIFF
--- a/src/api-docs.js
+++ b/src/api-docs.js
@@ -392,23 +392,8 @@ const apiDoc = {
                     description: 'The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.'
                   },
                   token: {
-                    type: 'object',
-                    required: ['uid', 'name', 'symbol'],
-                    description: 'Token to send the transaction, just in case is not HTR.',
-                    properties: {
-                      uid: {
-                        type: 'string',
-                        description: 'UID of the custom token to send the transaction.'
-                      },
-                      name: {
-                        type: 'string',
-                        description: 'Name of the custom token to send the transaction.'
-                      },
-                      symbol: {
-                        type: 'string',
-                        description: 'Symbol of the custom token to send the transaction.'
-                      },
-                    }
+                    type: 'string',
+                    description: 'Token of the output. If not sent, HTR will be chosen.'
                   },
                   'change_address': {
                     type: 'string',
@@ -422,11 +407,7 @@ const apiDoc = {
                   value: {
                     address: 'Wk2j7odPbC4Y98xKYBCFyNogxaRimU6BUj',
                     value: 100,
-                    token: {
-                      uid: '006e18f3c303892076a12e68b5c9c30afe9a96a528f0f3385898001858f9c35d',
-                      name: 'Test Coin',
-                      symbol: 'TSC'
-                    }
+                    token: '006e18f3c303892076a12e68b5c9c30afe9a96a528f0f3385898001858f9c35d'
                   }
                 }
               }
@@ -493,6 +474,7 @@ const apiDoc = {
                     type: 'array',
                     items: {
                       type: 'object',
+                      required: ['address', 'value'],
                       properties: {
                         address: {
                           type: 'string',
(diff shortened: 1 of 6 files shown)


src/api-docs.js
```diff
@@ -392,23 +392,8 @@ const apiDoc = {
...
-                        type: 'string',
-                        description: 'Name of the custom token to send the transaction.'
-                      },
-                      symbol: {
-                        type: 'string',
-                        description: 'Symbol of the custom token to send the transaction.'
-                      },
-                    }
+                    type: 'string',
+                    description: 'Token of the output. If not sent, HTR will be chosen.'
```
msbrogli, line 396: Suggestion: `'Token id of the output...'`.

src/api-docs.js
```diff
@@ -502,6 +484,10 @@ const apiDoc = {
                           type: 'integer',
                           description: 'The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.'
                         },
+                        token: {
+                          type: 'string',
+                          description: 'Token of the output. If not sent, HTR will be chosen.'
```
msbrogli, line 489: Same here.

src/index.js
```diff
@@ -258,7 +258,7 @@ walletRouter.get(
   const wallet = req.wallet;
   // Expects token uid
   const token = req.query.token || null;
-  const balance = wallet.getBalance(token);
+  const balance = wallet._getBalanceRaw(token);
```
msbrogli, line 261: It seems dangerous to use internal methods that we can freely change without considering compatibility. Besides that, I feel that not all implementation of the facade have this method available.

src/index.js
```diff
@@ -316,7 +317,7 @@ walletRouter.get('/address-index',
 walletRouter.get('/addresses', (req, res) => {
   const wallet = req.wallet;
   // TODO Add pagination
-  const addresses = wallet.getAllAddresses();
+  const addresses = wallet._getAllAddressesRaw();
```
msbrogli, line 320: Same here.

src/index.js
```diff
@@ -334,7 +335,7 @@ walletRouter.get('/tx-history',
   }
   const wallet = req.wallet;
   const limit = req.query.limit || null;
-  const history = wallet.getTxHistory();
+  const history = wallet._getHistoryRaw();
```
msbrogli, line 338: Same here.

src/index.js
```diff
@@ -586,36 +569,14 @@ walletRouter.post('/send-tx',
       isString: true,
       optional: true
     },
-    token: {
```
msbrogli, line 589: I like this change, but we are changing the API of the `/send-tx`. It means all integrations using it must adapt their code. Should we create a new version of the API and mark this one as deprecated?

src/index.js
```diff
@@ -396,28 +397,11 @@ walletRouter.post('/simple-send-tx',
     },
     token: {
       in: ['body'],
-      isObject: true,
+      isString: true,
       optional: true,
-      custom: {
```
msbrogli, line 401: I like this change, but we are changing the API of the `/simple-send-tx`. It means all integrations using it must adapt their code. Should we create a new version of the API and mark this one as deprecated?

msbrogli (changes requested):

src/index.js
pedroferreira1: Do you mean to create `/send-tx-new`? Or `/v2/send-tx`?

I agree that the integrations must adapt their code but it's only if they want to upgrade the headless.

I would need to map the old parameters to the new one and have two different APIs. I guess it's fair if we decide that it's important to do it.

src/api-docs.js
```diff
@@ -392,23 +392,8 @@ const apiDoc = {
...
-                        type: 'string',
-                        description: 'Name of the custom token to send the transaction.'
-                      },
-                      symbol: {
-                        type: 'string',
-                        description: 'Symbol of the custom token to send the transaction.'
-                      },
-                    }
+                    type: 'string | object',
+                    description: 'Token to send the transaction, just in case is not HTR. This parameter accepts a string with the token_uid (preferably) or the old format which is the full token object with {uid, name, symbol}. Support for the full token object is deprecated and will be removed soon.',
```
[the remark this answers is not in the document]
pedroferreira1: What do you think about this docs? This was the best way I found to have both parameters and I think it's very clear. 
image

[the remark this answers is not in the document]
msbrogli: Great! Thanks!

PR #119, opened 2021-08-20
feat: add new API to create NFT

pedroferreira1:
### Acceptance criteria

- There must be a new API to create an NFT.

### TODO

- [x] Still missing bump o library version that will be released after the merge of this PR (https://github.com/HathorNetwork/hathor-wallet-lib/pull/244)

### DIFF
--- a/src/api-docs.js
+++ b/src/api-docs.js
@@ -939,6 +939,109 @@ const apiDoc = {
         },
       },
     },
+    '/wallet/create-nft': {
+      post: {
+        summary: 'Create an NFT.',
+        parameters: [
+          {
+            name: 'x-wallet-id',
+            'in': 'header',
+            description: 'Define the key of the corresponding wallet it will be executed the request.',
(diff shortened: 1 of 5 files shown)


src/api-docs.js
```diff
@@ -939,6 +939,110 @@ const apiDoc = {
         },
       },
     },
+    '/wallet/create-nft': {
```
r4mmer, line 942: wouldn't a `POST` or `PUT` on `/wallet/nft` already indicate a creation?

pedroferreira1: I agree with you, it's redundant. But I think we should keep it anyway to keep the standard because we already have `/wallet/create-token`. What do you think?

r4mmer: I think that there are other paths that could be more clean and easier to understand. But we should keep the standard until we decide we need a refactor.

r4mmer (approved):

Issue #114, opened 2021-08-16
feat: Allow configuring which TxMiningService will be used by wallet-headless


luislhl:
Currently we have no way of passing some configuration to the service to change the TxMiningService used by it.

This also depends on changes in the hathor-wallet-lib project, so I also created a issue there: https://github.com/HathorNetwork/hathor-wallet-lib/issues/235

PR #131, opened 2021-10-05
feat: Allow setting tx mining url

luislhl:
closes: https://github.com/HathorNetwork/hathor-wallet-headless/issues/114

### Acceptance Criteria
- It should be possible to configure the TxMiningUrl to be used by the wallet
- The /status api should display the server url and network (it was broken)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1108,9 +1108,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.25.0",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.25.0.tgz",
-      "integrity": "sha512-1+LpSUY3fPrItqkba/CvxD8cn7r0SlZJvehxwi3f7IVrIkt77SBY0FVNVdnWCxP41TeXvizGTaNBkkZfuMEYqQ==",
+      "version": "0.25.1",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -10,15 +10,15 @@
       "integrity": "sha512-y5AohgeVhU+wO5kU1WGMLdocFj83xCxVjsVFa2ilII8NEwmBZvx7Ambq621FbFIK68loYJ9p43nfoi6es+rzSA==",
```
luislhl, line 10: We should try generating this with npm v5, to avoid changing so many lines.

config.js.template
```diff
@@ -10,6 +10,9 @@ module.exports = {
   network: 'mainnet',
   server: 'https://node2.mainnet.hathor.network/v1a/',
 
+  // Tx Mining Service
```
pedroferreira1, line 13: I think we can say that it's an optional config and if not set it will use the default from the lib. What do you think?

package.json
```diff
@@ -9,7 +9,7 @@
     "@babel/node": "^7.13.0",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@babel/preset-env": "^7.13.5",
-    "@hathor/wallet-lib": "^0.23.0",
+    "@hathor/wallet-lib": "^0.25.0",
```
luislhl, line 12: We should wait for the new version of wallet-lib, including a fix for a bug in 0.25.0

PR #149, opened 2022-02-21
Integration Tests: balance, tx-history, address-info and transaction routes

tuliomir:
This pull request is the first phase of the implementation of Integration Tests on the Hathor Wallet Headless, and lays the foundation for easily expanding the test scenarios in the future.

In short, on each PR the Github Actions were configured to run a `docker-compose` starting a Hathor private network, as described in [the private network guide](https://github.com/HathorNetwork/rfcs/blob/master/text/0033-private-network-guide.md), and then runs a series of tests on this network, validating the integration between the wallet and the network as a whole.

### Acceptance Criteria
- Run the integration tests on the following routes:
    - `/wallet/transaction`
    - `/wallet/address-info`
    - `/wallet/tx-history`
    - `/wallet/balance`
- The tests must be run on every PR and commit made on `dev` and `master` branches
    
### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

Issue #41, opened 2021-01-19
Use eslint to lint the code


andreabadesso:

tuliomir:
This issue is being addressed on #174

tuliomir:
Implemented on v0.12.0

PR #152, opened 2022-03-08
Integration tests: simple send-tx, send-tx, create-tokens, mint and melt tokens

tuliomir:
This pull request is the second phase of the implementation of Integration Tests on the Hathow Wallet Headless, and expands the scope of #149 .

#### Changes on previous tests:
- The wallets are now instantiated all at once on the `WalletHelper.startMultipleWalletsForTest()` method. This is a performance improvement, since wallet initialization is the most time-consuming process on the integration tests.
- Genesis wallet initialization was removed from the `setupTests-integration.js:beforeAll()` method to use the process above.

An effort was made to make the tests use as little transactions as possible, since transactions are the second slowest process on the tests. As the integration tests use Jest's `--runInBand` configuration, it became possible to use one test's resulting condition as precondition to the next tests.

Sometimes this may compromise the readability of the testing code, but it was a tradeoff for performance.

### Acceptance Criteria
- Run the integration tests on the following routes:
  - `/wallet/simple-send`
  - `/wallet/send-tx`
  - `/wallet/create-tokens`
  - `/wallet/mint-tokens`
  - `/wallet/melt-tokens`

### Security Checklist

### DIFF
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -57,10 +57,8 @@ describe('create token', () => {
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
-    expect(response.status)
-      .toBe(400);
-    expect(response.body.success)
-      .toBe(false);
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
     done();
   });
 
@@ -73,10 +71,8 @@ describe('create token', () => {
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
-    expect(response.status)
-      .toBe(400);
-    expect(response.body.success)
-      .toBe(false);
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
     done();
   });
 
@@ -92,12 +88,9 @@ describe('create token', () => {
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
-    expect(response.status)
-      .toBe(200);
-    expect(response.body.success)
-      .toBe(false);
-    expect(response.body.hash)
-      .toBeUndefined();
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.body.hash).toBeUndefined();
     done();
   });
 
@@ -111,14 +104,10 @@ describe('create token', () => {
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
-    expect(response.status)
-      .toBe(200);
-    expect(response.body.success)
-      .toBe(false);
-    expect(response.body.hash)
-      .toBeUndefined();
-    expect(response.body.error)
-      .toContain('Invalid token name');
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.body.hash).toBeUndefined();
+    expect(response.body.error).toContain('Invalid token name');
     done();
   });
 
@@ -132,10 +121,8 @@ describe('create token', () => {
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
-    expect(response.body.success)
-      .toBe(true);
-    expect(response.body.hash)
-      .toBeDefined();
+    expect(response.body.success).toBe(true);
+    expect(response.body.hash).toBeDefined();
     done();
   });
 
@@ -152,12 +139,10 @@ describe('create token', () => {
       .set({ 'x-wallet-id': wallet1.walletId });
 
     const transaction = response.body;
-    expect(transaction.success)
-      .toBe(true);
+    expect(transaction.success).toBe(true);
 
     const addr8 = await wallet1.getAddressInfo(9, transaction.hash);
-    expect(addr8.total_amount_received)
-      .toBe(amountTokens);
+    expect(addr8.total_amount_received).toBe(amountTokens);
     done();
   });
 
@@ -173,8 +158,7 @@ describe('create token', () => {
       .set({ 'x-wallet-id': wallet2.walletId });
 
     const transaction = response.body;
-    expect(transaction.success)
-      .toBe(true);
+    expect(transaction.success).toBe(true);
     const customTokenOutputIndex = TestUtils.getOutputIndexFromTx(transaction, 100);
 
     // If the custom token output is 0, the HTR will be on the output index 1. And vice-versa.
@@ -182,8 +166,7 @@ describe('create token', () => {
     const htrChange = transaction.outputs[htrOutputIndex].value;
 
     const addr8 = await wallet1.getAddressInfo(5);
-    expect(addr8.total_amount_received)
-      .toBe(htrChange);
+    expect(addr8.total_amount_received).toBe(htrChange);
     done();
   });
 });
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -1,4 +1,4 @@
-import { getRandomInt, TestUtils } from './utils/test-utils-integration';
+import { getRandomInt, TestUtils, WALLET_CONSTANTS } from './utils/test-utils-integration';
 import { WalletHelper } from './utils/wallet-helper';
 
 describe('create token', () => {
@@ -8,11 +8,6 @@ describe('create token', () => {
   const tokenA = {
     name: 'Token A',
     symbol: 'TKA',
-    uid: null
-  };
-  const tokenB = {
-    name: 'Token B',
-    symbol: 'TKB',
     uid: null
   };
 
@@ -28,6 +23,7 @@ describe('create token', () => {
 
   afterAll(async () => {
     await wallet1.stop();
+    await wallet2.stop();
   });
 
   // Testing failures first, that do not cause side-effects on the blockchain
@@ -57,6 +53,77 @@ describe('create token', () => {
 
     expect(response.status).toBe(400);
     expect(response.body.success).toBe(false);
+    done();
+  });
+
+  it('should reject an invalid destination address', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: tokenA.name,
+        symbol: tokenA.symbol,
+        amount: 1000,
+        address: 'invalidAddress'
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.body.error).toContain('base58');
+    done();
+  });
+
+  it('should reject an invalid change address', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: tokenA.name,
+        symbol: tokenA.symbol,
+        amount: 500,
+        address: await wallet1.getAddressAt(0),
+        change_address: 'invalidAddress'
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.body.error).toContain('Change address');
+    done();
+  });
+
+  // The application is incorrectly allowing to create a token for an outside address.
+  it.skip('should reject creating token for address not in the wallet', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: tokenA.name,
+        symbol: tokenA.symbol,
+        amount: 500,
+        address: WALLET_CONSTANTS.genesis.addresses[3]
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    done();
+  });
+
+  // The application is incorrectly allowing external addresses to receive the change
+  it.skip('should reject creating token for change address not in the wallet', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: tokenA.name,
+        symbol: tokenA.symbol,
+        amount: 500,
+        address: await wallet1.getAddressAt(0),
+        change_address: WALLET_CONSTANTS.genesis.addresses[3]
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.text).toContain('wallet');
     done();
   });
 
@@ -95,7 +162,7 @@ describe('create token', () => {
     done();
   });
 
-  it('should create a token successfully', async done => {
+  it('should create a token with only required parameters', async done => {
     const response = await TestUtils.request
       .post('/wallet/create-token')
       .send({
@@ -107,6 +174,11 @@ describe('create token', () => {
 
     expect(response.body.success).toBe(true);
     expect(response.body.hash).toBeDefined();
+
+    await TestUtils.pauseForWsUpdate();
+
+    const tkaBalance = await wallet1.getBalance(response.body.hash);
+    expect(tkaBalance.available).toBe(100);
     done();
   });
 
@@ -115,8 +187,8 @@ describe('create token', () => {
     const response = await TestUtils.request
       .post('/wallet/create-token')
       .send({
-        name: tokenB.name,
-        symbol: tokenB.symbol,
+        name: 'Token B',
+        symbol: 'TKB',
         amount: amountTokens,
         address: await wallet1.getAddressAt(9)
       })
@@ -127,8 +199,8 @@ describe('create token', () => {
 
     await TestUtils.pauseForWsUpdate();
 
-    const addr8 = await wallet1.getAddressInfo(9, transaction.hash);
-    expect(addr8.total_amount_received).toBe(amountTokens);
+    const addr9 = await wallet1.getAddressInfo(9, transaction.hash);
+    expect(addr9.total_amount_received).toBe(amountTokens);
     done();
   });
 
@@ -136,8 +208,8 @@ describe('create token', () => {
     const response = await TestUtils.request
       .post('/wallet/create-token')
       .send({
-        name: tokenB.name,
-        symbol: tokenB.symbol,
+        name: 'Token C',
+        symbol: 'TKC',
         amount: 100,
         change_address: await wallet2.getAddressAt(5)
       })
@@ -152,8 +224,36 @@ describe('create token', () => {
 
     await TestUtils.pauseForWsUpdate();
 
-    const addr8 = await wallet1.getAddressInfo(5);
-    expect(addr8.total_amount_received).toBe(htrChange);
+    const addr5 = await wallet2.getAddressInfo(5);
+    expect(addr5.total_amount_received).toBe(htrChange);
+    done();
+  });
+
+  it('should create a token with all available inputs', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: 'Token D',
+        symbol: 'TKD',
+        amount: 200,
+        address: await wallet2.getAddressAt(4),
+        change_address: await wallet2.getAddressAt(4)
+      })
+      .set({ 'x-wallet-id': wallet2.walletId });
+
+    const transaction = response.body;
+    expect(transaction.success).toBe(true);
+
+    // The only output with token_data equals zero is the one containing the HTR change
+    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0)
+    const htrChange = transaction.outputs[htrOutputIndex].value;
+
+    await TestUtils.pauseForWsUpdate();
+
+    const addr4 = await wallet2.getAddressInfo(4);
+    expect(addr4.total_amount_received).toBe(htrChange);
+    const addr4C = await wallet2.getAddressInfo(4, transaction.hash)
+    expect(addr4C.total_amount_available).toBe(200);
     done();
   });
 });
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -219,7 +219,7 @@ describe('create token', () => {
     expect(transaction.success).toBe(true);
 
     // The only output with token_data equals zero is the one containing the HTR change
-    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0)
+    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0);
     const htrChange = transaction.outputs[htrOutputIndex].value;
 
     await TestUtils.pauseForWsUpdate();
@@ -245,14 +245,14 @@ describe('create token', () => {
     expect(transaction.success).toBe(true);
 
     // The only output with token_data equals zero is the one containing the HTR change
-    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0)
+    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0);
     const htrChange = transaction.outputs[htrOutputIndex].value;
 
     await TestUtils.pauseForWsUpdate();
 
     const addr4 = await wallet2.getAddressInfo(4);
     expect(addr4.total_amount_received).toBe(htrChange);
-    const addr4C = await wallet2.getAddressInfo(4, transaction.hash)
+    const addr4C = await wallet2.getAddressInfo(4, transaction.hash);
     expect(addr4C.total_amount_available).toBe(200);
     done();
   });
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -5,15 +5,6 @@ describe('create token', () => {
   let wallet1;
   let wallet2;
 
-  const fundTx1 = {
-    hash: null,
-    index: null
-  };
-  const fundTx2 = {
-    hash: null,
-    index: null
-  };
-
   const tokenA = {
     name: 'Token A',
     symbol: 'TKA',
@@ -30,14 +21,8 @@ describe('create token', () => {
     wallet2 = new WalletHelper('create-token-2');
 
     await WalletHelper.startMultipleWalletsForTest([wallet1, wallet2]);
-    const fundTxObj1 = await wallet1.injectFunds(10, 0);
-    fundTx1.hash = fundTxObj1.hash;
-    fundTx1.index = TestUtils.getOutputIndexFromTx(fundTxObj1, 10);
-
-    const fundTxObj2 = await wallet1.injectFunds(10, 1);
-    fundTx2.hash = fundTxObj2.hash;
-    fundTx2.index = TestUtils.getOutputIndexFromTx(fundTxObj2, 10);
-
+    await wallet1.injectFunds(10, 0);
+    await wallet1.injectFunds(10, 1);
     await wallet2.injectFunds(10, 0);
   });
 
@@ -160,10 +145,9 @@ describe('create token', () => {
 
     const transaction = response.body;
     expect(transaction.success).toBe(true);
-    const customTokenOutputIndex = TestUtils.getOutputIndexFromTx(transaction, 100);
 
-    // If the custom token output is 0, the HTR will be on the output index 1. And vice-versa.
-    const htrOutputIndex = customTokenOutputIndex === 1 ? 0 : 1;
+    // The only output with token_data equals zero is the one containing the HTR change
+    const htrOutputIndex = transaction.outputs.findIndex(o => o.token_data === 0)
     const htrChange = transaction.outputs[htrOutputIndex].value;
 
     await TestUtils.pauseForWsUpdate();
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -191,8 +191,10 @@ describe('create token', () => {
 
     await TestUtils.pauseForWsUpdate();
 
+    const htrBalance = await wallet1.getBalance();
     const tkaBalance = await wallet1.getBalance(response.body.hash);
-    expect(tkaBalance.available).toBe(100);
+    expect(htrBalance.available).toBe(19); // The initial 20 minus 1
+    expect(tkaBalance.available).toBe(100); // The newly minted TKA tokens
     done();
   });
 
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -56,13 +56,13 @@ describe('create token', () => {
     done();
   });
 
-  it('should reject a name with more than 30 characters', async done => {
+  it.skip('should reject a name with more than 30 characters', async done => {
     const response = await TestUtils.request
       .post('/wallet/create-token')
       .send({
         name: 'Name input with more than 30 characters',
         symbol: tokenA.symbol,
-        amount: 1000
+        amount: 2000
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
@@ -78,7 +78,7 @@ describe('create token', () => {
       .send({
         name: tokenA.name,
         symbol: 'TKABCD',
-        amount: 1000
+        amount: 2000
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -141,9 +141,9 @@ describe('create token', () => {
     done();
   });
 
-  // Insuficcient funds
+  // insufficient funds
 
-  it('should reject for insuficcient funds', async done => {
+  it('should reject for insufficient funds', async done => {
     const response = await TestUtils.request
       .post('/wallet/create-token')
       .send({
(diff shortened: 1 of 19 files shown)


__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+    done();
+  });
+
+  // Lastly, testing success cases, which have side-effects
+
+  it('should send with only the output address and value', async done => {
+    const tx = await wallet1.sendTx({
+      fullObject: {
+        outputs: [{
+          address: await wallet2.getAddressAt(0),
```
pedroferreira1, line 340: Just like I said in the other files, I feel we should always check that the amount was sent to the correct address. When using address and when using change address. This applies to all tests below as well

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+      token: tokenA,
+    };
+    const tx = await wallet1.sendTx({
+      fullObject: sendOptions
+    });
+
+    expect(tx.success).toBe(true);
+    expect(tx.hash).toBeDefined();
+
+    // All 2000 TKA are now on Wallet2
```
pedroferreira1, line 891: We should assert that

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+      }],
+      token: tokenA,
+    };
+
+    const tx = await wallet2.sendTx({
+      fullObject: sendOptions
+    });
+
+    expect(tx.success).toBe(true);
+    expect(tx.hash).toBeDefined();
```
pedroferreira1, line 910: Should we assert wallet1 balance

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+          },
+          {
+            hash: tokenB.uid,
+            index: 1
+          },
+        ],
+        outputs: [
+          {
+            address: await wallet3.getAddressAt(7),
+            value: 1000,
```
pedroferreira1, line 990: Is it the full amount of utxos? We could try to use less than the full amount, this test will be a really good one. We will use selected utxos for tokenB, the lib will select utxos for HTR, then we will have change outputs for tokenB and maybe for HTR, will be really good

__tests__/integration/tx-history.test.js
```diff
@@ -24,16 +24,18 @@ describe('tx-history routes', () => {
...
-
       // A wallet with 5 transactions containing 10, 20, 30, 40 and 50 HTR each
       wallet2 = new WalletHelper('txHistory2');
-      await wallet2.start();
+
+      await WalletHelper.startMultipleWalletsForTest([wallet1, wallet2]);
+
       for (let amount = 10; amount < 60; amount += 10) {
-        const fundTx = await wallet2.injectFunds(amount, 1);
+        const fundTx = await wallet2.injectFunds(amount, 1, { doNotWait: true });
```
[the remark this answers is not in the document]
tuliomir: After making _a hundred_ tests I actually found out that waiting for the transaction to be updated via websocket is not very common and was adding unnecessary delays on many places. So I removed this option and, by default, there are no more delays on any transaction. The tests run time dropped by about 30%.

In order for the codes within the tests to remain self-explanatory, I created the `TestUtils.pauseForWsUpdate()` that does just that one second delay and keeps the test code clear. Is that a good solution?

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -140,12 +142,10 @@ export class TestUtils {
 
     // Wait until the wallet is actually started
     let status;
-    while (true) {
-      const res = await request
-        .get('/wallet/status')
-        .set(TestUtils.generateHeader(walletObj.walletId));
-      if (res.body?.statusCode === HathorWallet.READY) {
-        status = res.body;
+    while (!options.skipWait) {
```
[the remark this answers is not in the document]
tuliomir: The assumption here was that parameters given to the function are read-only and would never change.
This code was destructured for clarity to:
```js
if (options.waitForValidation) {
  while (true) {
    ...
  }
}
```

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -181,6 +195,23 @@ export class TestUtils {
...
 
+  /**
+   * Get the next addresses on this wallet that do not have transactions.
+   * The amount of addresses is the current gap limit.
+   * @param {string} walletId
+   * @returns {Promise<string[]>}
+   */
+  static async getSomeAddresses(walletId) {
+    const response = await TestUtils.request
+      .get('/wallet/addresses')
```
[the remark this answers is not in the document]
tuliomir: Updated JSDocs with this information.

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -229,4 +260,30 @@ export class TestUtils {
...
+  static getOutputIndexFromTx(transaction, value) {
+    if (!transaction?.outputs?.length) {
+      return null;
+    }
+
+    for (const index in transaction.outputs) {
+      if (transaction.outputs[index].value !== value) {
+        continue;
+      }
+      return +index;
```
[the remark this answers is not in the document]
tuliomir: It's a way to be succinct, but `parseInt` works just as well :+1: 
Actually, the linter demanded that `parteInt(index, 10)` was used, making it explicit that the conversion was for the decimal base.

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -229,4 +260,30 @@ export class TestUtils {
 
     return transaction;
   }
+
+  /**
+   * Retrieves the index containing the desired output value.
```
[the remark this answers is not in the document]
tuliomir: Updated JSDocs with this info.

__tests__/integration/utils/wallet-helper.js
```diff
@@ -39,20 +53,72 @@ export class WalletHelper {
...
+
+    // Requests the start of all the wallets in quick succession
+    for (const wallet of walletsArr) {
+      await TestUtils.startWallet({
+        walletId: wallet.walletId,
+        words: wallet.words,
+      }, {
+        skipWait: true
+      });
+      walletsPendingStart[wallet.walletId] = wallet;
```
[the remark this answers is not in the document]
tuliomir: Ok! Changed the variable name.

__tests__/integration/utils/wallet-helper.js
```diff
@@ -108,6 +176,44 @@ export class WalletHelper {
...
+      .set({ 'x-wallet-id': this.#walletId });
+
+    // An error happened
+    if (response.status !== 200 || response.body.success !== true) {
+      throw new Error(`Failure on /wallet/address-info: ${response.text}`);
+    }
+
+    // Returning explicitly each property to help with code completion / test writing
+    const addrInfo = response.body;
+    return {
```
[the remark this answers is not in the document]
tuliomir: For the javascript interpreter it's the same.

For the developer writing the tests, if using an IDE, the code completion works better. On each `expect(addr.` , all the relevant properties are readily shown for easy selection.

Should we decide to use _typescript_ sometime in the future, this would be unnecessary, but for these tests it proved to be a nice writing helper.

__tests__/integration/utils/wallet-helper.js
```diff
@@ -168,4 +281,86 @@ export class WalletHelper {
...
+   * @example
+   * wallet.sendTx({
+   *   destination: 'abc123',
+   *   value: 100,
+   *   token: 'def456'
+   * })
+   * @see https://wallet-headless.docs.hathor.network/#/paths/~1wallet~1simple-send-tx/post
+   * @param options
+   * @param {unknown} [options.fullObject] Advanced usage: a full body to send to post on 'send-tx'
+   * @param {{hash:string,index:number,token?:string}[]} [options.inputs] Optional Inputs
```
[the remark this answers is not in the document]
tuliomir: I expected a user trying to make a more specific use case to use the advanced `fullObject` parameter. But for documentation's sake to a developer learning the code, it's nice to have this information here too.

Two `@typedef`s were added describing inputs and outputs with examples.

If we ever change this method's contract, this will be one more place to update the documentation, though.

__tests__/integration/create-token.test.js
```diff
@@ -0,0 +1,172 @@
...
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.body.hash).toBeUndefined();
+    done();
+  });
+
+  it('should not create a token with the reserved HTR symbol', async done => {
```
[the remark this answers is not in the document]
tuliomir: Tests created but the one with the symbol name did not work as intended, so it's skipped for now.

__tests__/integration/melt-tokens.test.js
```diff
@@ -0,0 +1,122 @@
...
+    expect(response.body.success).toBe(true);
+    done();
+  });
+
+  it('should melt all the tokens', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/melt-tokens')
+      .send({
+        token: tokenA.uid,
+        address: await wallet1.getAddressAt(1),
```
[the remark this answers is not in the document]
tuliomir: I was following [the docs](https://wallet-headless.docs.hathor.network/#/paths/~1wallet~1melt-tokens/post) while making these tests and they do not mention these properties, so no tests were being made on them. It would be good to update these docs when possible.

Looking at the source code, I understood the actual working of this route and created new tests for the `change_address` and `deposit_address`. Also implemented assertions on the addresses' balances, following the suggested pattern on other tests.

__tests__/integration/create-token.test.js
```diff
@@ -0,0 +1,172 @@
...
+  it('should create a token successfully', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: tokenA.name,
+        symbol: tokenA.symbol,
+        amount: 100
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
```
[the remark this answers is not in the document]
tuliomir: Implemented the tests and had to skip them, since they're not working at the moment.
Opened issue #158 for that point.

__tests__/integration/mint-tokens.test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  // Success
+
+  it('should mint without a change address', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/mint-tokens')
+      .send({
+        token: tokenA.uid,
+        address: await wallet1.getAddressAt(1),
```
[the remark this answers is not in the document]
tuliomir: Balance validation assertions were made on all tests :+1:

__tests__/integration/mint-tokens.test.js
```diff
@@ -0,0 +1,154 @@
...
+    expect(response.body.success).toBe(true);
+    done();
+  });
+
+  it('should mint with a change address', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/mint-tokens')
+      .send({
+        token: tokenA.uid,
+        address: await wallet1.getAddressAt(1),
```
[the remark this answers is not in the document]
tuliomir: Implemented this assertion here and on some other opportunities found.

__tests__/integration/simple-send-tx.test.js
```diff
@@ -0,0 +1,295 @@
...
+  });
+
+  // Testing all transaction failures first, to have a easier starting test scenario
+
+  it('should not allow a transaction with an invalid value', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/simple-send-tx')
+      .send({
+        address: await wallet2.getAddressAt(0),
+        value: 'invalidValue',
```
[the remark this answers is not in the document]
tuliomir: Implemented :+1:

__tests__/integration/simple-send-tx.test.js
```diff
@@ -0,0 +1,295 @@
...
+    done();
+  });
+
+  // Executing all successful transactions
+
+  it('should make a successful transaction', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/simple-send-tx')
+      .send({
+        address: await wallet2.getAddressAt(0),
```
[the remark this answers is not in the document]
tuliomir: A new test was created to assert the change address must be on the same wallet.
Also, balance validations were implemented on every success test.

__tests__/integration/simple-send-tx.test.js
```diff
@@ -0,0 +1,295 @@
...
+    expect(response.body.error).toContain('Insufficient');
+
+    done();
+  });
+
+  it('should should make a successful transaction', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/simple-send-tx')
+      .send({
+        address: await wallet4.getAddressAt(0),
```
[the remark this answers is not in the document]
tuliomir: Done.

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+
+    expect(response.status).toBe(400);
+    expect(response.body.hash).toBeUndefined();
+    expect(response.body.success).toBe(false);
+    expect(response.body).toHaveProperty('error');
+    expect(response.body.error[0].msg).toContain('Invalid');
+    done();
+  });
+
+  it('should reject zero value', async done => {
```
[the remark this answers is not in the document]
tuliomir: Implemented

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+    done();
+  });
+
+  it('should reject an invalid input', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/send-tx')
+      .send({
+        inputs: [{
+          hash: 'invalidInput',
+          index: 0
```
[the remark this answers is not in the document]
tuliomir: Implemented this test. Worked the same as an "invalid input".

__tests__/integration/send-tx.test.js
tuliomir: Done!

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+
+  it('should send with only the filterAddress', async done => {
+    // Waiting for transactions to settle and spending the 20 HTR tokens above back to wallet1
+    await TestUtils.delay(1000);
+
+    const tx = await wallet2.sendTx({
+      fullObject: {
+        inputs: [{
+          type: 'query',
+          filter_address: await wallet2.getAddressAt(0)
```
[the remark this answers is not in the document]
tuliomir: New assertions were made to make sure the funds came from this address.

__tests__/integration/send-tx.test.js
tuliomir: Done

__tests__/integration/send-tx.test.js
tuliomir: Done

__tests__/integration/send-tx.test.js
tuliomir: A complex test was implemented on [1846a50](https://github.com/HathorNetwork/hathor-wallet-headless/pull/152/commits/1846a5009722e29a4965e8f5909cfb91e0ba6d90) for this purpose.

Can you check if it covers what you have thought of here?

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1005 @@
...
+    const sendOptions = {
+      inputs: [{
+        hash: tkaTx1.hash,
+        index: 0
+      }], // Using index 0 of main transaction
+      outputs: [{
+        address: await wallet2.getAddressAt(0),
+        value: 200
+      }],
+      token: tokenA,
```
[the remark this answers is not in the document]
tuliomir: I tried testing both apis, but ended up not covering the no-input and single-input cases on the new api. Multiple input and output cases are tested already.

New tests were added on [bca626b](https://github.com/HathorNetwork/hathor-wallet-headless/pull/152/commits/bca626b2f42c9df29a311a7d8081be5a5cdf8f73) to cover these cases.

__tests__/integration/create-token.test.js
```diff
@@ -0,0 +1,290 @@
...
+        amount: 1000
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.body.success).toBe(false);
+    expect(response.body.error).toContain('maximum size');
+    done();
+  });
+
+  // The result is an error with the message "maximum size", but consumes the funds. Must be fixed.
```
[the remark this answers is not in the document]
tuliomir: On every debug I made, there were no other transactions executed on the fullnode. So, part of the test works as intended.
However, every transaction made after it triggers a "Insuficcient Funds" response.

This leads me to conclude that there is an error on the internal wallet-lib storage that happens and, since there are no websocket updates because the transaction did not happen, this mistake is not corrected.

As we plan to open an issue solving every skipped test on this suite, I understand that this bug is already mapped.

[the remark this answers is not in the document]
pedroferreira1: I've just checked the code and you are correct. It's a local control and a bug. An issue should be created to fix this (as you said, we will go through all skipped tests (when are we going to do this?)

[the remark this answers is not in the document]
pedroferreira1: Looking at the code, the same problem should happen with the name, are you sure it's not?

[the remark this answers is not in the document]
tuliomir: You are right. I increased the "name" test amount to consume all the funds and the other tests started failing.

Probably the reason the error was not being triggered before is because at least one small transaction was being made successfully before the bugged funds were consumed, which updated the balance via websocket and "hid" the error.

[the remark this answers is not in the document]
pedroferreira1: So we should skip it as well and in the issue explain that both tests refer to the same problem

__tests__/integration/mint-tokens.test.js
```diff
@@ -0,0 +1,252 @@
...
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
+    expect(response.text).toContain('amount');
+    done();
+  });
+
+  // The application is allowing minting for an address outside the wallet
```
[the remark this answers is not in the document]
tuliomir: Test removed

__tests__/integration/melt-tokens.test.js
```diff
@@ -0,0 +1,342 @@
...
+    expect(addr5htr.total_amount_available).toBe(1);
+    expect(addr5tka.total_amount_available).toBe(0);
+
+    const balance1 = await wallet1.getBalance(tokenA.uid);
+    expect(balance1.available).toBe(500 - 100);
+
+    const balance1htr = await wallet1.getBalance();
+    const balance1tka = await wallet1.getBalance(tokenA.uid);
+    expect(balance1htr.available).toBe(2 + 3 + 1);
+    expect(balance1tka.available).toBe(800 - 300 - 100);
```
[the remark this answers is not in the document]
tuliomir: Removed unnecessary assertion. It was a leftover from previous iterations.

__tests__/integration/melt-tokens.test.js
```diff
@@ -0,0 +1,342 @@
...
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    expect(response.text).toContain('invalid');
+    done();
+  });
+
+  // The application is incorrectly allowing a deposit address outside the wallet
```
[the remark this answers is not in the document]
tuliomir: Test removed

__tests__/integration/create-token.test.js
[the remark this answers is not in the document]
tuliomir: Understood.
The test was updated with some assertions for specific `address` balances and general `wallet` balances, to ensure these criteria.

__tests__/integration/create-token.test.js
```diff
@@ -0,0 +1,290 @@
...
+        address: WALLET_CONSTANTS.genesis.addresses[3]
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+    done();
+  });
+
+  // The application is incorrectly allowing external addresses to receive the change
```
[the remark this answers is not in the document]
andreabadesso: Why is this a bug? Shouldn't we allow the user to build the transaction anyway he wants to?

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -72,11 +74,14 @@ export class TestUtils {
...
+   * Whenever the tests need to check for a wallet/address balance, there should be a small pause
+   * to allow for the Fullnode's Websocket connection to update the Wallet Headless' local caches.
+   *
+   * The delay period here should be optimized for this purpose.
+   * @returns {Promise<void>}
    */
-  static get request() {
-    return request;
+  static async pauseForWsUpdate() {
+    await TestUtils.delay(1000);
```
andreabadesso, line 84: Maybe this should be a env variable

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -181,16 +200,85 @@ export class TestUtils {
     return response.body.address;
   }
 
+  /**
```
andreabadesso, line 203: **Comment unrelated to line number**

Maybe we should create an issue to move all those methods from the `TestUtils` that communicate with the headless API to a new library, maybe a `headless-client` library

What do you think?

__tests__/integration/utils/wallet-helper.js
```diff
@@ -27,6 +27,20 @@ export class WalletHelper {
...
   #started = false;
 
+  /**
+   * Creates a wallet object but does not start it on server
+   * @param {string} walletId
+   * @param {string} [words] 24 words
+   */
+  constructor(walletId, words) {
+    if (!walletId) {
+      throw new Error(`Wallet must have a walletId`);
```
andreabadesso, line 37: Use single quotes

__tests__/integration/utils/wallet-helper.js
```diff
@@ -39,30 +53,87 @@ export class WalletHelper {
...
-    this.#words = words || TestUtils.generateWords();
+    // If the genesis wallet is not instantiated, start it. It should be always available
+    const { genesis } = WALLET_CONSTANTS;
+    const isGenesisStarted = await TestUtils.isWalletReady(genesis.walletId);
+    if (!isGenesisStarted) {
+      walletsArr.unshift(new WalletHelper(genesis.walletId, genesis.words));
+    }
+
+    // Requests the start of all the wallets in quick succession
+    for (const wallet of walletsArr) {
```
andreabadesso, line 85: What happens if we do this in parallel?

__tests__/integration/utils/wallet-helper.js
```diff
@@ -39,30 +53,87 @@ export class WalletHelper {
...
+        walletId: wallet.walletId,
+        words: wallet.words,
+      });
+      walletsPendingReady[wallet.walletId] = wallet;
+    }
+
+    // Enters the loop checking each wallet for its status
+    while (true) {
+      const pendingWalletIds = Object.keys(walletsPendingReady);
+      if (!pendingWalletIds.length) {
```
andreabadesso, line 96: Maybe we should also add a timeout on this loop

I guess that if the wallet headless loses connection to the full-node, it will return not ready indefinitely

__tests__/integration/utils/wallet-helper.js
```diff
@@ -82,6 +153,8 @@ export class WalletHelper {
     };
   }
 
+  __setStarted() { this.#started = true; }
```
andreabadesso, line 156: Out of curiosity, why two `_`s?

__tests__/integration/utils/wallet-helper.js
```diff
@@ -162,10 +264,105 @@ export class WalletHelper {
...
+    // Logs the results
+    const metadata = {
+      originWallet: this.#walletId,
+      hash: transaction.hash,
+      ...sendOptions
+    };
+    if (options.destinationWallet) {
+      metadata.destinationWallet = options.destinationWallet;
+    }
+    await loggers.test.insertLineToLog(`Transferring funds`, metadata);
```
andreabadesso, line 356: Use single quotes

setupTests-integration.js
```diff
@@ -21,11 +20,8 @@ beforeAll(async () => {
   const testLogger = new TxLogger(testName);
   testLogger.init();
   loggers.test = testLogger;
-
-  await TestUtils.startWallet(WALLET_CONSTANTS.genesis);
 });
 
 // This function will run after each test file is executed
 afterAll(async () => {
```
andreabadesso, line 26: Remove `afterAll`?

__tests__/integration/melt-tokens.test.js
```diff
@@ -0,0 +1,342 @@
...
+    done();
+  });
+
+  it('should melt with change address only', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/melt-tokens')
+      .send({
+        token: tokenA.uid,
+        amount: 100,
+        change_address: await wallet1.getAddressAt(7),
```
andreabadesso, line 238: I don't like the fact that we need to be careful with the tests order and the address indexes we are using

Not sure what is the best way to fix it tho

What do you think about having a global counter on this file, increasing it on each test, so you can do something like `wallet1.getAddressAt(addressCounter.getNextValue())`?

```javascript
const addressCounter = {
  value: 0,
  getNextValue: function() {
    this.value = this.value + 1;
    return this.value;
  },
};
```

__tests__/integration/melt-tokens.test.js
```diff
@@ -0,0 +1,342 @@
...
+      .send({
+        token: 'invalidToken',
+        amount: 100
+      })
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(200);
+    expect(response.body.success).toBe(false);
+
+    // Even though the result is correct, the error thrown is not related. Should be fixed later.
```
andreabadesso, line 53: Add `TODO` to the comment so we can find it using grep later

__tests__/integration/send-tx.test.js
```diff
@@ -0,0 +1,1361 @@
...
+      .set({ 'x-wallet-id': wallet1.walletId });
+
+    expect(response.status).toBe(400);
+    expect(response.text).toContain('Invalid');
+    expect(response.text).toContain('value');
+    done();
+  });
+
+  // Insuficcient funds
+  it('should reject for insuficcient funds', async done => {
```
andreabadesso, line 250:
```sed
s/insuficcient/insufficient/g
```

andreabadesso (changes requested):

__tests__/integration/create-token.test.js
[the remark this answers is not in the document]
pedroferreira1: Change address by definition is the amount that returns to the same wallet. Maybe we should allow that in the lib but not in the headless, to prevent human error.

__tests__/integration/utils/wallet-helper.js
tuliomir: That is a good refactoring, but had no noticeable result on the test times.

__tests__/integration/utils/wallet-helper.js
tuliomir: This is [another convention](https://www.thoughtco.com/and-in-javascript-2037515) for _private_ or _hidden internal methods_ on javascript.

Since I implemented the `startMultipleWalletsForTest`, I had to start a wallet from an outside context, but I didn't want this method to show up on code completion for developers. So, it's a kind of a _soft encapsulation_.

If it seems too weird for the current code style I can revert it to just `setStarted()`

__tests__/integration/utils/wallet-helper.js
tuliomir: Done

__tests__/integration/send-tx.test.js
tuliomir: :man_facepalming: fixed

__tests__/integration/utils/test-utils-integration.js
tuliomir: Moved it to the config file, together with the transaction log settings.

setupTests-integration.js
tuliomir: I kept it for easier reference, should anyone need to use it later.
But since there already is the `setupTests.js` file with this same reference code, it's better to remove it indeed. :+1:

__tests__/integration/utils/wallet-helper.js
tuliomir: Implemented a timeout option with a default to 60 seconds, adjustable via config file.
Also added some logging to help identify if any test suite makes this fail. It showed me that currently the wallet initializations are between 8 - 47 seconds.

__tests__/integration/utils/test-utils-integration.js
tuliomir: Seems like a great idea! Opened issue #162 for this discussion.

__tests__/integration/melt-tokens.test.js
tuliomir: Done :+1:

__tests__/integration/melt-tokens.test.js
tuliomir: I was thinking on this direction when decided to implement some caches within the `wallet` class. I didn't go very far in this direction because increasing complexity on the test code could lead to bugs on the tests themselves.

I see two ways to lessen this problem:
### Declare clear variables containing the addresses and their purposes
For example, on the closure of `describe('melt tokens')`:
```js
const wallet1Addresses = {
      initialFunds: {
        index: 0,
        address: 'WSA9tePoHSvJ2ktVGCnVfefKxTwQJizAis'
      },
      basicDestination: {
        index: 1,
        address: 'WYE86DaZcdhibA3XBXnqaXGNsGK2T4Va9W'
      },
      meltDeposit1: {
        index: 2,
        address: 'WhLfSp44WHr5m1GML6BdtpkeyDqjY6byid'
      },
      meltChange1: {
        index: 3,
        address: 'WUnTgUFXeFksQUJEfaN1c3NdyErez3mvrw'
      }
    };
```

### Making actual independent tests
Each test could make its own funding transaction using `wallet.injectFunds()` , select addresses using `wallet.getNextAddress()` ( I could easily implement a way to "mark as used" ) store the starting balance on the relevant addresses and then execute the tests. Assertions would be made on values relative to the starting situation.

They would be more verbose, a bit slower to run, but would have the advantage to not depend on each other so much and ease maintainability.

What do you think of this tradeoff? Could we try one of these solutions on the next batch of tests, maybe on `filter-utxos` and `consolidate-utxos`?

__tests__/integration/utils/wallet-helper.js
pedroferreira1: The `/start` API does almost nothing, I guess that's the reason it didn't increase performance. But I agree it's a good refactor

config.js.template
```diff
@@ -40,6 +40,14 @@ module.exports = {
   },
   */
 
+  // Integration test configs
```
pedroferreira1, line 43: I don't think we should add anything in this file. This is the template for `src/config.js` and you are adding some test template for `__tests__/integration/configuration/config.js.template`.

We shouldn't even have a template for that, I think.

pedroferreira1 (changes requested):

config.js.template
tuliomir: The reason for this template is that on the integration tests a full instance of the Wallet Headless has to be started, with the correct configuration for the accompanying dockerized private network. So, the `config.js.template` must be present somehow for the GitHub workflow to move to `/src` folder and use it on CI.
https://github.com/HathorNetwork/hathor-wallet-headless/blob/a8151b2e1bdbad5e7ac0d509a40172c74736a8b8/.github/workflows/integration-test.yml#L34

If the main configuration file is not a good place for these settings to be, an option would be making a new file. Here follow my suggested change: moving the specific integration test configurations to [a `test.config.js` file](https://github.com/HathorNetwork/hathor-wallet-headless/commit/02d690b45c2e77f7c660a97d708b003ad3ba6952), on the git source control.

As one of the comments here was about changing the configurations via environment variables, I also added another feature to override this config with the environment. Having two config files _and_ environment variables to allow for an application to run seems a bit confusing to me, but maybe we have to overdo it a little in order to reach a final solution.

Could @pedroferreira1 and @andreabadesso validate this suggestion?

__tests__/integration/utils/wallet-helper.js
andreabadesso: As we've discussed on gather, we should use the wallet-service's eslint to keep consistency between projects

config.js.template
andreabadesso: Me too

pedroferreira1 (approved):

__tests__/integration/utils/wallet-helper.js
tuliomir: For now, while we don't implement the linter on #41 , I've changed the code manually to comply with this `single-quote` rule

andreabadesso (approved):

Issue #151, opened 2022-02-25
Issue when using filter_query and custom tokens


obiyankenobi:
On `/send-tx`, when we have a query filter, the UTXOs are selected according to the total sum of the outputs, whereas it should get the output sum for each token.

https://github.com/HathorNetwork/hathor-wallet-headless/blob/ac4a34c85f4327c3e49774f48075c0b48f42ba05/src/index.js#L705-L709

If there's a query filter, we sum the outputs, regardless of the token. We then proceed to get the UTXOs and just use the token in the query filter. It may happen that there's no token in the query filter but there's one in the output.

To sum up, there are 2 bugs:
1) We sum all outputs, regardless of tokens;
2) We get the UTXOs for just one token (by default HTR), even if there are multiple tokens in the outputs;

### Steps to reproduce:

1) Start a wallet and send 1 HTR to it. Balance is 1 HTR:

```
curl -X GET -H "X-Wallet-Id: id" http://localhost:8000/wallet/balance/ | jq
{
  "available": 100,
  "locked": 0
}
```

2) Create a token:

```
curl --location --request POST 'http://localhost:8000/wallet/create-token' \
--header 'X-Wallet-Id: id' \
--header 'Content-Type: application/json' \
--data-raw '{
  "name": "BugCoin",
  "symbol": "BUG",
  "amount": 10000
}'
```

It has balance: 
```
curl -X GET -H "X-Wallet-Id: id" http://localhost:8000/wallet/balance/?token=00000000f2822546538a71aa21540115f04eb0650d9f5af048f2e1a191519e2a | jq
{
  "available": 10000,
  "locked": 0
}
```

3) Get the address that has this token balance:

```
curl -X GET -H "X-Wallet-Id: id" http://localhost:8000/wallet/transaction/?id=00000000f2822546538a71aa21540115f04eb0650d9f5af048f2e1a191519e2a | jq
{
  "tx_id": "00000000f2822546538a71aa21540115f04eb0650d9f5af048f2e1a191519e2a",
  "version": 2,
  "weight": 8.000001,
  "timestamp": 1645814206,
  "is_voided": false,
  "inputs": [
    {
      "value": 100,
      "token_data": 0,
      "script": "dqkU5HCZxSjNv0aWajr5xxbXj7IHzFCIrA==",
      "decoded": {
        "type": "P2PKH",
        "address": "WjVuqJ9FYgK27wLarEfWNs1ADP87GqeCxQ",
[...]

PR #160, opened 2022-03-14
Fix issue when user filters UTXOs by address query and custom tokens

tuliomir:
The feature to `send-tx` querying by filter address was only working for HTR tokens, and would have wrong behaviors when any of the outputs had a custom token.

### Acceptance Criteria
- A transaction must be successful with input address filter query and custom tokens, as described in #151 

### Security Checklist

### DIFF
--- a/__tests__/integration/send-tx.test.js
+++ b/__tests__/integration/send-tx.test.js
@@ -1393,6 +1393,23 @@ describe('filter query + custom tokens', () => {
      */
   });
 
+  afterAll(async () => {
+    await wallet1.stop();
+  });
+
+  it('should reject for insufficient funds (custom)', async done => {
+    // Sending all the tokens to facilitate address-info validation
+    const txErr = await wallet1.sendTx({
+      inputs: [{ type: 'query', filter_address: await wallet1.getAddressAt(0) }],
+      outputs: [{ token: bugCoin.uid, address: await wallet1.getAddressAt(1), value: 2000 }],
+    }).catch(err => err.innerError);
+
+    expect(txErr.status).toBe(200);
+    expect(txErr.body.success).toBe(false);
+    expect(txErr.body.error).toContain('No utxos');
+    done();
+  });
+
   it('should send the custom token with a query filter by address 0', async done => {
     // Sending all the tokens to facilitate address-info validation
     const tx = await wallet1.sendTx({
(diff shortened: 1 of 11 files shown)


src/index.js
```diff
@@ -703,19 +731,36 @@ walletRouter.post('/send-tx',
...
+            success: false,
+            error: 'No utxos available for the query filter for this amount.',
+            token: tokenUid
+          };
+          res.send(response);
+          lock.unlock(lockTypes.SEND_TX);
+          return;
+        }
+
+        for (const utxo of utxos) {
```
r4mmer, line 758: Why this change? We map `utxos` to `treatedInputs` then assign it to `inputs`, the same as we did before.

__tests__/test-utils.js
```diff
@@ -47,6 +47,9 @@ class TestUtils {
     "Weg6WEncAEJs5qDbGUxcLTR3iycM3hrt4C",
     "WSVarF73e6UVccGwb44FvTtqFWsHQmjKCt",
     "Wc5YHn861241iLY42mFT8z1dT1UdsNWkfs",
+    "WU6KGNPuaRG4VCKC1fsKDh4fRZGbkqxG8Y",
```
r4mmer, line 50: Why these new addresses?

src/index.js
```diff
@@ -563,6 +571,13 @@ const getUtxosToFillTx = (wallet, sumOutputs, options) => {
   }
 }
 
+/**
+ * @typedef TokenOutput
```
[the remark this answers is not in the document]
tuliomir: This type is first declared on line 704 and used on 718:
https://github.com/HathorNetwork/hathor-wallet-headless/blob/ca9f1938522cfe8ee86e6f712e3d2224f53215a3/src/index.js#L718-L722

It is used as a helper object only for internal processes, with no interface to the clients.

src/index.js
tuliomir: Each token iterated may have multiple `utxos`. The `treatedInputs` variable is basically an intermediate to reduce them all to a single resulting array.

Only after all the output tokens have been iterated we overwrite the `inputs` variable with the final result. That was the most concise solution I found.
Do you have an alternative suggestion?

__tests__/test-utils.js
tuliomir: Since I had to use more of the wallet's addresses on the new tests, any query that involves the _gap limit_ also fetched more addresses that were not specified before on the mocked code.

For example, this `addresses` test:
https://github.com/HathorNetwork/hathor-wallet-headless/blob/ca9f1938522cfe8ee86e6f712e3d2224f53215a3/__tests__/addresses.test.js#L28

This is part of the refactoring I had to do on all the mocked tests fixtures on commit https://github.com/HathorNetwork/hathor-wallet-headless/pull/160/commits/e8f39488a88d573a63c581d6d9d8734f8fa1b08c

src/index.js
r4mmer: Didn't see that the inputs assignment was on another context. 
But i guess we could remove the `treatedInputs` anyway with:
```js
      // init inputs as empty array then
      inputs = inputs.concat(utxos.map((utxo) => {
        return {txId: utxo.tx_id, index: utxo.index};
      }));
```

src/index.js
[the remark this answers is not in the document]
r4mmer: I understood the usage, but shouldn't this comment be better near where it's used?
I think this would avoid confusion.

__tests__/integration/send-tx.test.js
```diff
@@ -1359,3 +1359,246 @@ describe('send tx (custom tokens)', () => {
...
+  });
+
+  afterAll(async () => {
+    await wallet1.stop();
+  });
+
+  it('should reject for insufficient funds (custom)', async done => {
+    // Sending all the tokens to facilitate address-info validation
+    const txErr = await wallet1.sendTx({
+      inputs: [{ type: 'query', filter_address: await wallet1.getAddressAt(0) }],
```
pedroferreira1, line 1403: I would just add a new error test case with amount 100 (an amount that the wallet has) but with an address without the utxo, e.g. the address 1.

pedroferreira1 (approved):

Issue #172, opened 2022-03-31
Refactor the express application


tuliomir:
Break the `index.js` file into routes, controllers and services, as the convention for Express applications.

tuliomir:
Implemented on v0.12.0

PR #174, opened 2022-04-01
Refactor of ExpressJS files

tuliomir:
This PR aims to refactor this Express application into a more conventional format of separating files into routes, controllers and services.

### Acceptance Criteria
- The application should be structured in routes, controllers and services
- Solves issue #172
- Solves issue #41 

### Risks and compromises
At the current state of the **Integration Tests**, the test results are sometimes failing due to timeouts and other minor failures related to excessive CPU consumption.

There are also some issues with the code coverage, raised by lgtm-com, that existed before this refactor and are outside its scope to change.

On commit [91be24f](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/91be24f869030a5cd40ee705ec76894b6cf854c8) changes were made to the configurations to make most possible test pass. All tests not described below were passing at the time of that commit, which indicates the refactor was successful in not changing the application logic.

These changes were reverted on [f628571](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/f6285716bdff6e61c89d3a380dd337393cb3d393) to ensure this PR is within its scope.

#### Failing Tests
1) **Multisig** - send tx (HTR) › Should send a transaction with minimum signatures
2) **Multisig** - send tx (HTR) › Should send a transaction with max signatures

#### Risks of merging with these failures
Both failures are still being investigated. But since the multisig feature is being blocked on the `constants` file, these failures pose no immediate risk to the users should this code reach production. Their errors have been pinpointed on the [test results for 91be24f](https://github.com/HathorNetwork/hathor-wallet-headless/runs/5842656374?check_suite_focus=true) to be these:
- `{"success":false,"error":"OP_CHECKSIG: failed\nOP_CHECKSIG: failed\nOP_CHECKSIG: failed\nOP_CHECKSIG: failed\nStack left with more than one value"}`
- `{"success":false,"error":"OP_DUP: empty stack"}`

### Security Checklist
- [ ] Understand the risks of the failing tests and the need to make a separate task to fix them in the near future.


package.json
```diff
@@ -24,6 +24,7 @@
     "test_network_down": "docker-compose -f ./__tests__/integration/docker-compose.yml down",
     "dev": "nodemon --exec babel-node index.js",
     "start": "babel-node src/index.js",
+    "lint": "eslint . --fix",
```
luislhl, line 27: Shouldn't we run this in Github Actions?

tuliomir: Good idea! Done on [318dad2](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/318dad2802459a93004e60684240e0657e8dc702)

pedroferreira1 (changes requested):

src/controllers/index.controller.js
```diff
@@ -0,0 +1,224 @@
+const { walletUtils, errors, Connection, HathorWallet } = require('@hathor/wallet-lib');
```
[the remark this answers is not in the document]
tuliomir: Added headers to all application files on [2de3762](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/2de376233e37cd4bf2714b4954891eb6608b6e2b)

__tests__/test-utils.js
```diff
@@ -106,65 +111,69 @@ class TestUtils {
...
 
     const response = await request
-      .post("/start")
-      .send({ seedKey: seedKey, "wallet-id": walletId, multisig: multisig });
+      .post('/start')
+      .send({ seedKey, 'wallet-id': walletId, multisig });
 
     if (response.status !== 200) {
-      throw new Error("Unable to start the wallet");
+      throw new Error(response.text);
```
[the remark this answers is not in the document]
tuliomir: Fixed on [35d3c67](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/35d3c6710b812f48a1b1fb19f0bd35c9101be229)

__tests__/integration/multisig.test.js
```diff
@@ -33,29 +32,29 @@ describe('send tx (HTR)', () => {
...
-    'wh2eCGzUK9rLThr5D6tyCfckHpBjS97ERA',
-    'wZvajxVp3LabcZiY3XPrivrXiSS6wphRu7',
-    'wgPbL1WzbrEntepHRC92UX6mA2EmaqfDqt',
-    'wbdx4g3rucX3WHmZRXjPEKtRfZ7XSnCGKf',
-    'wiKTnqSN11ukuCWEXRVrRTTPo2mw4fGue3',
-    'wQ4aQP4YqJqfwshLggR2w1Gg3UFhhKhVKs',
-    'wca2xk9S2MVn2UrKh78UScdwXz3xrTp8Ky',
-    'wcUZ6J7t2B1s8bqRYiyuZAftcdCGRSiiau',
-  ];
+  // const addresses = [
```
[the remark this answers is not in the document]
tuliomir: Added an explanation to the comment block on [7ea530c](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/7ea530c291bb9e49ca7d9b04879ebd0582e448e2).

This variable needed to be removed because of the lint `no-unused-var` rule but, as you said, it's a good reference for developing tests. So instead of removing it, I thought better to keep it as a comment.

setupTests-integration.js
```diff
@@ -15,16 +17,24 @@ function getTestNameFromGlobalJasmineInstance() {
 
 // Mock config file
 jest.mock(
-  "./src/config",
-  () => require("./__tests__/integration/configuration/config-fixture.js"),
-  { virtual: true }
+  './src/config',
+  () => {
+    let config = require('./__tests__/integration/configuration/config-fixture');
```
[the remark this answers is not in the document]
tuliomir: After the application refactor, all tests ( integration and mocked ) stopped working because of this mock file code on the setup. Sometimes the retrieved config file was correct, but in other cases it was nested inside a `default` property on the required js.

I believe it is related to the config and constants being referred not only by the `index.js` file anymore, but from `js` files spread on various folders with different relative addresses. The simplest solution I found was to make this check for the `default` property on the required files.

Since this is directly related to the refactor of the application, I believe it belongs to this PR. Do you agree?

setupTests-integration.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable global-require */
 import * as Path from 'path';
```
r4mmer, line 2: Maybe import only `parse` since it's the only one used.

setupTests.js
```diff
@@ -1,25 +1,35 @@
-import TestUtils from "./__tests__/test-utils";
+/* eslint-disable import/no-extraneous-dependencies,global-require */
+import TestUtils from './__tests__/test-utils';
```
r4mmer, line 2: Why not remove this?

src/services/wallets.service.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// TODO: Make this a map instead of a common object. It's safer.
```
r4mmer, line 8: `TODO` on this PR or for later?

src/routes/wallet.routes.js
```diff
@@ -0,0 +1,380 @@
...
+  createNft
+);
+
+/**
+ * POST request to stop a wallet
+ * For the docs, see api-docs.js
+ */
+walletRouter.post('/stop', stop);
+
+walletRouter.use('/tx-proposal', txProposalRouter);
```
r4mmer, line 378: Maybe put this after the middleware on the start of the file.

r4mmer (changes requested):
I like the organization but some files were `src/<section>/<name>.<section>.js` and it felt repetitive.

Do you think we should remove the second `<section>`? Or is there any reason to keep it?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Install dependencies
+        run: npm install
+
+      # This file is required by most controllers and triggers "import/no-unresolved" if missing
+      - name: Copy config file
+        run: cp ./__tests__/integration/configuration/config-fixture.js ./src/config.js
+
+      - name: Lint
+        run: npm run lint
```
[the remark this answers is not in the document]
tuliomir: Since our tests are currently not green, I tried to create the minimum possible interference with them. So I created a new workflow _yml_.

But now that you mention, I see our [wallet mobile](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/master/.github/workflows/main.yml) follows this practice of running the lint within the `main` workflow.

I agree that's a good idea, but I would rather make all tests green before making this move.
What do you think?

.github/workflows/integration-test.yml
```diff
@@ -6,10 +6,10 @@ on:
       - dev
     tags:
       - v*
-    pull_request:
```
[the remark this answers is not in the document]
tuliomir: I like the idea of only running the integration tests after the unit ones, but maybe not on the same workflow. 

Both tests have a timeout of 40 minutes and generate great amounts of logs. While we're trying to fine tune them, I think it's more helpful to keep them on isolated runs with separate outputs. The cost benefit may not be a relevant trade-off for the isolation of problems while in active development.

I think that with some optimizations on the headless, the tests and the underlying lib we could join those tests in the future. But while we don't get to that point, maybe other solutions like the [wait on check action](https://github.com/marketplace/actions/wait-on-check) or the [`workflow_run` property](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) could be good alternatives to this ordering of actions.

What do you think?

setupTests.js
tuliomir: These are _lint_ rules that I found it better to ignore for now, since fixing them could create confusion or take too long to investigate the proper way to implement them:

### [`import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md)
This accuses the `mock-socket` dependency to be not explicitly declared on the `package.json` dependencies. Looking at [the rules we follow from `airbnb-base`](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L68-L97), I see that the setup filename we've chosen is not within their convention.

At a future time we could move the setup files into the `__tests__` folder and this would be solved.

### [`global-require`](https://eslint.org/docs/rules/global-require)
On lines `18` and `29` we require javascript files within a very controlled clojure scope on Jest. If we imported these files into global variables and tried to use them in the mock `factory` method, we would get the following error:
```
babel-plugin-jest-hoist: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
```

src/services/wallets.service.js
tuliomir: To do later. This is a comment to reinforce this [`Use of externally-controlled format string` issue](https://lgtm.com/projects/g/HathorNetwork/hathor-wallet-headless/snapshot/ce8e08444593ddf2828ed6c4611edaeff4c3e528/files/src/controllers/index.controller.js?sort=name&dir=ASC&mode=heatmap#xb9f3e98ba428f368:1) raised by LGTM bot.

src/routes/wallet.routes.js
tuliomir: Done on [e0b7301](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/e0b7301ba3e5148be371bb786c9c622919049655)

setupTests-integration.js
tuliomir: Good point. Done on [e0b7301](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/e0b7301ba3e5148be371bb786c9c622919049655)

src/routes/wallet.routes.js
```diff
@@ -0,0 +1,371 @@
...
+  getAddress
+);
+
+/**
+ * GET request to get an address index
+ * For the docs, see api-docs.js
+ */
+walletRouter.get(
+  '/address-index',
+  query('address').isString(),
```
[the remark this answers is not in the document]
pedroferreira1: Can you create an issue for these tests that don't exist?

setupTests-integration.js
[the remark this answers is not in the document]
pedroferreira1: Yes, I agree.

pedroferreira1 (approved):

setupTests-integration.js
pedroferreira1: I guess the idea with this import change is that you would be able to remove the comment above that disables the lint check. `/* eslint-disable global-require */`

I think you should remove it now that you've changed the import.

.github/workflows/lint.yml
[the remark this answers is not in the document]
luislhl: I think the interference caused by linters is not that big. We will be able to differentiate failures caused by them or by the tests.

setupTests-integration.js
tuliomir: Actually this `global-require` is there because of lines 22 and 33, and it's still required.

This issue is being discussed on [a thread below](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174#discussion_r845464136) about the other setup file, but the reasons are the same.

tuliomir:
@r4mmer about both the file name and the folder containing the `<section>`, this is a convention I see on many articles and projects on ExpressJS.

I believe the idea is to help when the developer is working on multiple levels of the same part of the project. For example, instead of having three `wallet.js` files open, one would have the `wallet.routes.js`, `wallet.controller.js` and `wallet.service.js`, which removes ambiguity at a glance of the screen.

One example of this file structure is found [on this article](https://medium.com/codechef-vit/a-better-project-structure-with-express-and-node-js-c23abc2d736f) about ExpressJS project structures from 2020.

.github/workflows/lint.yml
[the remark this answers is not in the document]
tuliomir: I tried adding the linter to the main workflow on [a555448](https://github.com/HathorNetwork/hathor-wallet-headless/pull/174/commits/a555448eef4c6e42f73094ff903b1ba461e5ec83), but it broke when tried to run on Node 8.x, possibly also on 10.x

We could use additional actions or configurations to enable the linting only on specific node versions, but I think there is little benefit for this extra complexity.

Do you have any suggestions on this?

src/routes/wallet.routes.js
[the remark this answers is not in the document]
tuliomir: Created issue #176 - `Create tests for basic query routes`

r4mmer (approved):

PR #194, opened 2022-05-17
feat: add length max for data script outputs

pedroferreira1:
### Acceptance Criteria
- APIs that create data script outputs should have maximum length of data string. For now, we have only send-tx and create-nft.

### Security Checklist

### DIFF
--- a/src/routes/wallet/wallet.routes.js
+++ b/src/routes/wallet/wallet.routes.js
@@ -15,6 +15,7 @@ const {
   getAddressIndex
 } = require('../../controllers/wallet/wallet.controller');
 const txProposalRouter = require('./tx-proposal.routes');
+const { MAX_DATA_SCRIPT_LENGTH } = require('../../constants');
 
 const walletRouter = Router({ mergeParams: true });
 walletRouter.use(walletMiddleware);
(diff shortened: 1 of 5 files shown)


tuliomir (approved):

src/routes/wallet/wallet.routes.js
```diff
@@ -217,6 +218,11 @@ walletRouter.post(
     'outputs.*.data': {
       in: ['body'],
       isString: true,
+      isLength: {
+        options: {
+          max: MAX_DATA_SCRIPT_LENGTH
+        }
```
andreabadesso, line 224:
```suggestion
        options: {
          max: MAX_DATA_SCRIPT_LENGTH,
        },
```

andreabadesso (changes requested):

src/routes/wallet/wallet.routes.js
andreabadesso: Since our linter rules are not currently explicit about this situation, we've decided to move on with this PR and refactor the lint rules (and the files) in another PR

andreabadesso (approved):

Issue #191, opened 2022-05-04
fix: skipped multisig tests


tuliomir:
During #164 , two multisig integration tests failed. Since they were out of the scope of that PR, they were just skipped.
Durint #189, the mulsitsig test on the `start wallet` has also failed and was skipped.

All tests on the multisig mocked and integration suites must pass.

r4mmer:
All tests related to this issue are fixed and have been enabled to run during CI

PR #198, opened 2022-05-24
tests: do not skip multisig tests

r4mmer:
### Acceptance Criteria
Fix problems described on #191

### Security Checklist

### DIFF
--- a/__tests__/integration/multisig.test.js
+++ b/__tests__/integration/multisig.test.js
@@ -3,6 +3,7 @@ import { TestUtils } from './utils/test-utils-integration';
 import { WalletHelper } from './utils/wallet-helper';
 import { multisigWalletsData } from '../../scripts/helpers/wallet-precalculation.helper';
 import precalculatedMultisig from './configuration/precalculated-multisig-wallets.json';
+import { loggers } from './utils/logger.util';
 
 describe('send tx (HTR)', () => {
   let wallet1;
@@ -103,8 +104,8 @@ describe('send tx (HTR)', () => {
     const tx = {
       input: [fundTx1],
       outputs: [
-        { address: precalculatedMultisig[0].addresses[1], value: 100 },
-        { address: precalculatedMultisig[0].addresses[2], value: 270 },
+        { address: precalculatedMultisig[0].addresses[0], value: 100 },
+        { address: precalculatedMultisig[0].addresses[1], value: 270 },
       ],
     };
 
@@ -113,7 +114,7 @@ describe('send tx (HTR)', () => {
       .post('/wallet/tx-proposal')
       .send(tx)
       .set({ 'x-wallet-id': wallet1.walletId });
-    console.log(JSON.stringify(response.body));
+    loggers.test.insertLineToLog('multisig[should fail lt minsig]: proposal', { body: response.body });
 
     const { txHex } = response.body;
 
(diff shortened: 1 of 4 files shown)


src/controllers/wallet/tx-proposal.controller.js
```diff
@@ -117,9 +119,9 @@ async function signAndPush(req, res) {
   }
 
   const { txHex } = req.body;
-  const signatures = req.body.signatures || [];
+  const signatures = req.body.signatures.slice(0, req.wallet.multisig.minSignatures) || [];
```
r4mmer, line 122: The MultiSig implementation on the fullnode only checks the minimum signatures and rejects the transaction since the extra signatures remain on stack when validating.
This slice is meant to cut any signatures the user sends above the minimum so this error does not happen.

__tests__/integration/multisig.test.js
```diff
@@ -122,57 +123,14 @@ describe('send tx (HTR)', () => {
...
       .post('/wallet/tx-proposal/sign-and-push')
       .send({ txHex, signatures: [sig1, sig2] })
       .set({ 'x-wallet-id': wallet1.walletId });
-    console.log(JSON.stringify(response.body));
-
-    expect(response.status).toBe(200);
-    expect(response.body.success).toBe(false);
-  });
-
-  it('Should fail to send a transaction with more than max signatures', async () => {
```
r4mmer, line 137: This test was removed because the sign and push api now removes any excess signatures so this behavior would not fail anymore.

src/controllers/wallet/tx-proposal.controller.js
```diff
@@ -86,9 +86,10 @@ async function signTx(req, res) {
...
```
tuliomir, line 89: Trying to understand the behavior here: assuming `minSignatures` is 2 and there are 5 signatures at maximum.

If a user receives the `txHex` with 2 signatures and runs it through this `signTx` method, it will just add another one, to a total of 3 signatures.
If a user receives the `txHex` with 3 signatures and runs it through `signTx`, it will slice off one of the signatures and add its own, to a total of 3 signatures again.

Is this the desired behavior? I feel like it would want to slice _after_ the `assemblePartialTransaction`, and not before.
If this is not, I believe a test specific for this case would be a good idea.

tuliomir (changes requested):

tuliomir (approved):

src/controllers/wallet/tx-proposal.controller.js
```diff
@@ -70,6 +70,22 @@ async function getMySignatures(req, res) {
   }
 }
 
+function assemblePartialTransaction(wallet, txHex, signatures) {
```
pedroferreira1, line 73: Missing docstring

__tests__/integration/multisig.test.js
pedroferreira1: This was changed, so should we add this test back?

pedroferreira1: Actually I've seen that you already have a test with more than minSignatures, so should be okay to remove this test, right?

pedroferreira1 (approved):

src/controllers/wallet/tx-proposal.controller.js
r4mmer: Added on [75da341](https://github.com/HathorNetwork/hathor-wallet-headless/pull/198/commits/75da341be33cdddd02804f4e146d06912b25527b)

tuliomir (approved):

PR #199, opened 2022-05-26
tests: individual wallets for each test

r4mmer:
### Summary

This fixes the intermittent errors on the start tests (described here #191) 

### Acceptance Criteria
- Each tests suite should have it's own individual wallet id so they can all be independent.

### Security Checklist

### DIFF
--- a/__tests__/address-info.test.js
+++ b/__tests__/address-info.test.js
@@ -1,11 +1,21 @@
 import TestUtils from './test-utils';
 
+const walletId = 'stub_address_info';
+
 describe('address-info api', () => {
+  beforeAll(async () => {
+    await TestUtils.startWallet({ walletId });
+  });
+
+  afterAll(async () => {
+    await TestUtils.stopWallet({ walletId });
+  });
+
   it('should return 200 with a valid body', async () => {
     const response = await TestUtils.request
       .get('/wallet/address-info')
       .query({ address: 'WewDeXWyvHP7jJTs7tjLoQfoB72LLxJQqN' })
-      .set({ 'x-wallet-id': TestUtils.walletId });
+      .set({ 'x-wallet-id': walletId });
     expect(response.status).toBe(200);
     expect(response.body.success).toBeTruthy();
     expect(response.body.index).toBe(0);
@@ -15,7 +25,7 @@ describe('address-info api', () => {
     const response = await TestUtils.request
       .get('/wallet/address-info')
       .query({ address: 'WewDeXWyvHP7jJTs7tjLoQfoB72LLxJQqN', token: '01' })
-      .set({ 'x-wallet-id': TestUtils.walletId });
+      .set({ 'x-wallet-id': walletId });
     expect(response.status).toBe(200);
     expect(response.body.success).toBeTruthy();
     expect(response.body.index).toBe(0);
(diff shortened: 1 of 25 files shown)


__tests__/test-utils.js
```diff
@@ -179,6 +181,18 @@ class TestUtils {
...
+    TestUtils.logger = winston.createLogger({
+      level: 'silly',
+      transports: [
+        new winston.transports.Console({
+          format: winston.format.simple(),
+          level: 'silly',
+        }),
+      ],
+    });
+  }
```
r4mmer, line 195: This is a simple version of the logger of the integration tests, this was made because jest has some issues with console outputs within the test methods which we usually use for debugging.

__tests__/start.test.js
```diff
@@ -8,21 +8,25 @@ import { WALLET_CONSTANTS } from './integration/configuration/test-constants';
  * leaked connections at the end of the test.
  */
 
+const walletId = 'stub_wallet_start';
+
 describe('start api', () => {
-  beforeEach(async () => TestUtils.stopWallet());
+  beforeEach(async () => {
```
r4mmer, line 14: Here we use `beforeEach` instead of the usual `beforeAll`/`afterAll` because we are testing the start wallet api

__tests__/address-info.test.js
```diff
@@ -1,11 +1,21 @@
...
+const walletId = 'stub_address_info';
+
 describe('address-info api', () => {
+  beforeAll(async () => {
+    await TestUtils.startWallet({ walletId });
+  });
+
+  afterAll(async () => {
+    await TestUtils.stopWallet({ walletId });
+  });
```
r4mmer, line 12: Any test that uses a wallet should start and cleanup it's own wallet. The words used for all wallets will be the same but the `walletId` should be unique to each test.

__tests__/test-utils.js
```diff
@@ -103,16 +117,62 @@ class TestUtils {
...
+      }
+      if (res.body?.statusCode === HathorWallet.ERROR) {
+        throw new Error(res.body?.message);
+      }
+      if (exitIfClosed && res.body?.statusCode === HathorWallet.CLOSED) {
+        return false;
+      }
+      await new Promise(resolve => {
+        setTimeout(resolve, 500);
+      });
```
tuliomir, line 153: We have a `delay` function on [core.util.js](https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/__tests__/integration/utils/core.util.js#L8-L17) that I think would be better than calling `setTimeout` here, for simplicity.

Maybe we should open a new issue later to move the `core.util` file to the main tests folder, instead of a "integration test" specific one. It's a small change, but a good one to keep things tidy.

r4mmer: I saw the delay method but I wanted to keep the same execution as the `startWallet`. Maybe on another PR we can make an actual utility core to be used on both integration and unit tests.

__tests__/balance.test.js
```diff
@@ -1,10 +1,20 @@
 import TestUtils from './test-utils';
 
+const walletId = 'stub_balance';
+
 describe('balance api', () => {
+  beforeAll(async () => {
+    await TestUtils.startWallet({ walletId, preCalculatedAddresses: TestUtils.addresses });
```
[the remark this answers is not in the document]
r4mmer: I didn't use the precalculated addresses where the address generation process is tested. Do you think it would be better to have this explicitly commented on the tests?

tuliomir (approved):

pedroferreira1 (approved):

PR #200, opened 2022-06-01
docs: testing instructions

tuliomir:
### Acceptance Criteria
- There should be a new markdown document with information about the testing environment

### Security Checklist


TESTING.md
```diff
@@ -0,0 +1,88 @@
...
+```sh
+npm run test
+```
+
+To run the integration tests:
+```sh
+npm run test_integration
+```
+
+The data that is used on static code/coverage analysis on GitHub come from the mocked context. So, when adding new code it's important to add tests on the _mocked context_ to cover it. As we see further on this document, this is because the _integration tests_ do not need to cover all edge cases of the code.
```
pedroferreira1, line 16: > this is because the _integration tests_ do not need to cover all edge cases of the code.

I don't like this sentence, looks like the integration tests is just for one code example and nothing else. We run tests for error parameters, different parameters and different scenarios, so I guess we should try to have for all edge cases of the code, even in the integration tests.

tuliomir: Agreed that this sentence is not very helpful. Removed in [cc50b19](https://github.com/HathorNetwork/hathor-wallet-headless/pull/200/commits/cc50b19cb368fd4ea5c9e95c0b81ccb5c593ff52).

The responsibilites of each type of test will be discussed at a later moment. For now, this document only needs to inform and help devs/testers to understand and run this application's tests.

pedroferreira1 (approved):

PR #211, opened 2022-06-21
feat: atomic swap

r4mmer:
### Acceptance Criteria
- Implement the atomic-swap endpoints described [here](https://github.com/HathorNetwork/hathor-wallet/issues/308)

### Security Checklist


src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    res.send({ success: true, signatures: signatures.serialize() });
+  } catch (err) {
+    res.send({ success: false, error: err.message });
+  }
+}
+
+async function getLockedUTXOs(req, res) {
+  try {
```
pedroferreira1, line 237: I would create the same `SEND_TX` lock here because this `selected_as_input` parameter might be affected if a tx is being sent in the moment

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+}
+
+async function unlockInputs(req, res) {
+  const validationResult = parametersValidation(req);
+  if (!validationResult.success) {
+    res.status(400).json(validationResult);
+    return;
+  }
+
```
pedroferreira1, line 267: I would create the same `SEND_TX` lock here because this `selected_as_input` parameter might be affected if a tx is being sent in the moment

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+    const response = await sendTransaction.runFromMining();
+    res.send({ success: true, ...mapTxReturn(response) });
+  } catch (err) {
+    res.send({ success: false, error: err.message });
+  } finally {
+    lock.unlock(lockTypes.SEND_TX);
+  }
+}
+
+async function getInputData(req, res) {
```
pedroferreira1, line 207: We don't have any tests for this API

__tests__/atomic-swap/get-my-signatures.test.js
```diff
@@ -0,0 +1,76 @@
...
+    // cleanup mock
+    spy.mockRestore();
+    spyValidate.mockRestore();
+  });
+
+  afterEach(() => {
+    spy.mockClear();
+  });
+
+  it('should fail if partial_tx is not a string', async () => {
```
pedroferreira1, line 43: Maybe add also a test without this parameter

__tests__/atomic-swap/locked-utxos.test.js
```diff
@@ -0,0 +1,89 @@
...
+          new ProposalInput('hash1', 1, 1, 0),
+          new ProposalInput('hash3', 1, 1, 0),
+        ];
+        return partialTx;
+      });
+    const spyMark = jest.spyOn(hathorLib.HathorWallet.prototype, 'markUtxoSelected').mockImplementation(() => {});
+
+    const response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal/unlock')
+      .send({ partial_tx: 'partial-tx-data' })
```
pedroferreira1, line 74: We should test the 400 error without the parameter

__tests__/atomic-swap/sign-and-push.test.js
```diff
@@ -0,0 +1,127 @@
...
+    // cleanup mock
+    spy.mockRestore();
+    spyValidate.mockRestore();
+  });
+
+  afterEach(() => {
+    spy.mockClear();
+  });
+
+  it('should fail if partial_tx is not a string', async () => {
```
pedroferreira1, line 44: What if we don't send the required parameters?

__tests__/atomic-swap/sign.test.js
```diff
@@ -0,0 +1,154 @@
+import hathorLib from '@hathor/wallet-lib';
+import { ProposalInput, ProposalOutput, PartialTx } from '@hathor/wallet-lib/lib/models/partial_tx';
+import TestUtils from '../test-utils';
+
+const walletId = 'stub_atomic_swap_tx_proposal_sign';
+
+describe('tx-proposal sign api', () => {
```
pedroferreira1, line 7: This test file is exactly the same as the `sign-and-push` file. We should reuse all the methods only changing the API, so we don't have this duplicated code

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,259 @@
...
+    - If not present, will create a new proposal.
+- change_address
+    - Use this address for change where needed (e.g. sending 10 HTR but the UTXO has 15, a change output will be generated)
+    - If not present, an address from the caller wallet will be used.
+- send_tokens
+  - Will add inputs with the amount of tokens specified, may add a change output if needed.
+  - An array of objects with:
+    - value
+    - token: If not present will default to 00 (HTR)
+- receive_tokens
```
pedroferreira1, line 79: It's not really clear the different between `receive_tokens` and `outputs`. They have exactly the same parameters

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,259 @@
...
+```bash
+$ curl  -X POST \
+        -H "X-Wallet-Id: 123" \
+        -H "Content-type: application/json" \
+        --data "{\"partial_tx\":\"<proposalString>\"}" \
+        http://localhost:8000/wallet/atomic-swap/tx-proposal/unlock
+{
+  "success": true,
+}
+```
```
pedroferreira1, line 259: I like the description above but I feel we should add a new section with a more detailed and verbose example.

Bob and Alice would like to exchange Token A (TKNA) and HTR. They agree that Bob will send 10 HTR and receive 2 TKNA. Bob will then start the atomic swap creating the tx-proposal, (tx proposal code to send 10 HTR and receive 2 TKNA), get the serialized data, sends to Alice, Alice decodes, validates it's correct, adds it's own output and fills the inputs, checks tx is completed and correct, get her signatures, send back to Bob the full tx with her signatures. Bob decodes the tx, validates, add his own signatures, push to the network.

But with verbose details on each step. I would also add a second example with one of the atomic swap wallets being a multisig one, I think we need this in more details rather than just saying it's possible and briefly mentioning how to do it.

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+    expect(response.body.success).toBe(true);
+
+    await TestUtils.pauseForWsUpdate();
+  });
+
+  it('should exchange custom tokens between 2 wallets', async () => {
+    // wallet1 proposes the transaction
+    let response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal')
+      .send({ send_tokens: [{ token: tokenTx1.hash, value: 10 }] })
```
pedroferreira1, line 106: I would use the second test to test with `outputs` and `inputs` instead of `send_tokens` and `receive_tokens`

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+    await wallet1.stop();
+    await wallet2.stop();
+    await walletMultisig.stop();
+  });
+
+  it('should exchange HTR between 2 wallets', async () => {
+    // wallet1 proposes the transaction
+    let response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal')
+      .send({ send_tokens: [{ value: 10 }] })
```
pedroferreira1, line 60: Thinking about the logical flow in my opinion, the one who proposes the transaction will say that he wants to send 10 HTR and receive 10 HTR, so we can fill the utxos and the outputs, right?

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+      .post('/wallet/atomic-swap/tx-proposal/sign-and-push')
+      .send({ partial_tx: data, signatures: [signature] })
+      .set({ 'x-wallet-id': wallet2.walletId });
+    loggers.test.insertLineToLog('atomic-swap[HTR P2PKH]: push', { body: response.body });
+    expect(response.body.success).toBe(true);
+
+    await TestUtils.pauseForWsUpdate();
+  });
+
+  it('should exchange custom tokens between 2 wallets', async () => {
```
pedroferreira1, line 102: I don't understand what's happening in this example. wallet1 is sending 10 `tokenTx1` and wallet2 is receiving 10 `tokenTx1`? This is a simple send, right?

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+
+    await TestUtils.pauseForWsUpdate();
+  });
+
+  it('should exchange multiple tokens between 2 wallets', async () => {
+    // wallet1 proposes the transaction
+    let response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal')
+      .send({
+        send_tokens: [{ token: tokenTx1.hash, value: 5 }, { token: '00', value: 10 }],
```
pedroferreira1, line 153: This is a nice example, 2 tokens in exchange of 1

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+          { token: tokenTx1.hash, value: 50 },
+        ],
+      })
+      .set({ 'x-wallet-id': walletMultisig.walletId });
+    loggers.test.insertLineToLog('atomic-swap[P2SH]: proposal', { body: response.body });
+    expect(response.body.isComplete).toBe(false);
+
+    let { data } = response.body;
+
+    // wallet1 updates the proposal
```
pedroferreira1, line 218: What happens here? It's important to describe in the comment
Will the proposal be completed substituted? Or it will merge somehow?

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,259 @@
...
+      }
+    ]
+  }
+}
+```
+
+## How to: negotiation phase
+
+This phase consists of participants creating a proposal which will be serialized to a string which will be passed among participants.
+Each participan will add the tokens they want to send and receive from the transaction, once the transaction is complete the participants can move to the signing phase.
```
tuliomir, line 54:
```suggestion
Each participant will add the tokens they want to send and receive from the transaction, once the transaction is complete the participants can move to the signing phase.
```

tuliomir (changes requested):
A few suggestions about tests, that may affect many files:

**Error handling tests**
I would suggest expanding the controller tests to cover the failure cases. For example, the functions `PartialTxProposal.fromPartialTx` and `PartialTxInputData.signData` can both throw in case of a business logic error, and it would be important to assert that the end-user will have the correct response JSON on those cases.

**Untested code branches**
The recently-added `codecov` plugin is indicating some of the controller's lines are not being covered by the tests. Some of those end up being good suggestions for new tests, that may reveal bugs we haven't found on the current code.

scripts/helpers/wallet-precalculation.helper.js
```diff
@@ -41,6 +41,21 @@ export const multisigWalletsData = {
 };
 multisigWalletsData.walletConfig.pubkeys = multisigWalletsData.pubkeys;
 
+export const singleMultisigWalletData = {
```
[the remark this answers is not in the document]
r4mmer: It doesn't actually make sense, but a multisig with a single pubkey is good for testing some cases since we just need to init 1 wallet and get 1 signature.
We will not use this for the actual multisig tests because we need to test all cases which include multiple participants and less the the min participants.

src/schemas.js
```diff
@@ -0,0 +1,213 @@
...
+    optional: true,
+  },
+  'signatures.*': {
+    in: ['body'],
+    errorMessage: 'Invalid signature',
+    isString: true,
+  },
+};
+
+export const partialTxOrTxHexSchema = {
```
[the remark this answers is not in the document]
r4mmer: Changed to using `oneOf` and removed the `partialTxOrTxHexSchema`.

src/schemas.js
```diff
@@ -0,0 +1,213 @@
...
+  },
+};
+
+export const partialTxSignatureSchema = {
+  ...partialTxSchema,
+  signatures: {
+    in: ['body'],
+    errorMessage: 'Invalid signatures array',
+    isArray: true,
+    notEmpty: true,
```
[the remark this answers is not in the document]
r4mmer: I tested, if the parameter is not present no errors will raise (optional part) and if it exists, it must be an array and not empty.
So an empty array will return 400, an array with incorrect parameters will return 400 and not having the array at all will not return an error.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+  const sendTokens = req.body.send_tokens || [];
+  const receiveTokens = req.body.receive_tokens || [];
+  const changeAddress = req.body.change_address || null;
+  const partialTx = req.body.partial_tx || null;
+  let markAsSelected = true;
+  if (req.body.lock !== undefined) {
+    markAsSelected = req.body.lock;
+  }
+
+  if (inputs.length === 0
```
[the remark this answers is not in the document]
r4mmer: Are we going to only allow 1 operation per call?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+    res.status(400).json(validationResult);
+    return;
+  }
+
+  const network = req.wallet.getNetworkObject();
+  const partialTx = req.body.partial_tx;
+  const proposal = PartialTxProposal.fromPartialTx(partialTx, network);
+
+  try {
+    storage.setStore(req.wallet.store);
```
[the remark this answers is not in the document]
r4mmer: Done

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+
+  const network = req.wallet.getNetworkObject();
+
+  const partialTx = req.body.partial_tx;
+  const signatures = req.body.signatures || [];
+
+  const proposal = PartialTxProposal.fromPartialTx(partialTx, network);
+
+  try {
+    await proposal.signData('123');
```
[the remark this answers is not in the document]
r4mmer: Because this is the `sign-and-push` API, we can change it to just `push` and let all signatures come from `get-my-signatures`.
I think it would be a good option as well.

src/schemas.js
```diff
@@ -0,0 +1,213 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const { constants: { HATHOR_TOKEN_CONFIG } } = require('@hathor/wallet-lib');
```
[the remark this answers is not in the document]
r4mmer: Removed

src/schemas.js
```diff
@@ -0,0 +1,213 @@
...
+  'outputs.*.token': {
+    in: ['body'],
+    errorMessage: 'Invalid token uid',
+    isString: true,
+    optional: true,
+  },
+  'outputs.*.timelock': {
+    in: ['body'],
+    errorMessage: 'Invalid timelock',
+    isInt: true,
```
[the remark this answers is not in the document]
r4mmer: Done

src/schemas.js
```diff
@@ -0,0 +1,213 @@
...
+        // Test if txId is a 64 character hex string
+        if (!(/^[0-9a-fA-F]{64}$/.test(value))) return false;
+        return true;
+      }
+    },
+  },
+  'inputs.*.index': {
+    in: ['body'],
+    errorMessage: 'Invalid input index',
+    isInt: true,
```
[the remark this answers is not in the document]
r4mmer: Done

src/schemas.js
```diff
@@ -0,0 +1,213 @@
...
+      options: {
+        min: 1,
+      },
+    },
+    toInt: true,
+  },
+  'receive_tokens.*.timelock': {
+    in: ['body'],
+    errorMessage: 'Invalid timelock',
+    isInt: true,
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/atomic-swap/locked-utxos.test.js
r4mmer: Done.

__tests__/atomic-swap/sign-and-push.test.js
r4mmer: Added all invalid params cases to the tests.

__tests__/atomic-swap/sign.test.js
r4mmer: I'll wait the decision on changing the `sign-and-push` api to just `push` before changing this.

ATOMIC_SWAP.md
r4mmer: The only actual difference is that for `receive_tokens`, the address is optional and if not present will generate an address from the wallet.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: Done.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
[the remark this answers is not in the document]
pedroferreira1: It's not only a `push` API, you should receive all signatures, in my opinion, assemble the tx, mine and push, so it's a bit more than just push but I think we should change this

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: Created unit tests, the integration tests were already created.

__tests__/integration/atomic-swap.test.js
r4mmer: Yes. The the integration tests are covering:
- simple send: HTR exchange.
- simple send: custom token exchange.
- exchange multiple custom tokens and HTR.
- exchange multiple custom tokens and HTR between P2PKH and P2SH.

The simple send cases may not be an intended use of the atomic-swap but I think we should cover this case on the tests.

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+    // wallet2: sign data
+    response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal/get-my-signatures')
+      .send({ partial_tx: data })
+      .set({ 'x-wallet-id': wallet2.walletId });
+    loggers.test.insertLineToLog('atomic-swap[1TK P2PKH]: get signatures', { body: response.body });
+    expect(response.body).toEqual({
+      success: true,
+      isComplete: false,
+      signatures: expect.any(String),
```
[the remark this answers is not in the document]
r4mmer: It will return a string but when we run the deserialize method it will not contain any input data.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+    return;
+  }
+
+  const canStart = lock.lock(lockTypes.SEND_TX);
+  if (!canStart) {
+    res.send({ success: false, error: cantSendTxErrorMessage });
+    return;
+  }
+
+  const network = req.wallet.getNetworkObject();
```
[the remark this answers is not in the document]
r4mmer: Done.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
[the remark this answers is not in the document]
r4mmer: Changed the API to not sign with the loaded wallet and only assemble with the data it receives (and replicated this on the sign api).

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: I agree with tulio

__tests__/integration/atomic-swap.test.js
r4mmer: Changed the comments to reflect which actions are being added to the proposal.

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,259 @@
...
+1. Use the `/wallet/atomic-swap/tx-proposal` endpoint to get the proposal string and send to the participants.
+1. Each participant should update the proposal string using the `/wallet/atomic-swap/tx-proposal` endpoint and add their tokens and what they expect to receive.
+    - proposal strings cannot be merged, only updated, so each participant should take turns updating and sending the updated proposal string.
+1. The endpoint will return `isComplete: true` when the proposal string is ready to be signed.
+    - The proposal string can still be updated, but it will only be ready to sign again if the funds for inputs and outputs match.
+
+### POST /wallet/atomic-swap/tx-proposal
+
+Create a transaction proposal with multiple inputs and outputs.
+inputs, outputs, send_tokens, receive_tokens are operations which will add inputs and outputs to the partial_tx, can use one or many operations at once.
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/atomic-swap/get-my-signatures.test.js
```diff
@@ -0,0 +1,76 @@
...
+    const response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal/get-my-signatures')
+      .send({ partial_tx: 123 })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBeFalsy();
+  });
+
+  it('should return the signatures for the inputs we own on the transaction', async () => {
```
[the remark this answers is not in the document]
r4mmer: Done

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,292 @@
...
+    res.status(400).json({ success: false, error: ['Should have at least one operation'] });
+    return;
+  }
+
+  const proposal = partialTx
+    ? PartialTxProposal.fromPartialTx(partialTx, network) : new PartialTxProposal(network);
+
+  for (const input of inputs) {
+    let error = null;
+    txApi.getTransaction(input.txId, data => {
```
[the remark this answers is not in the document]
r4mmer: Changed the promise handling.

__tests__/atomic-swap/sign-and-push.test.js
```diff
@@ -0,0 +1,127 @@
...
+    const proposal = new hathorLib.PartialTxProposal(testnet);
+    proposal.partialTx = partialTx;
+    return proposal;
+  };
+  const scriptFromAddress = base58 => {
+    const p2pkh = new hathorLib.P2PKH(new hathorLib.Address(base58, { network: testnet }));
+    return p2pkh.createScript();
+  };
+
+  const spy = jest.spyOn(hathorLib.PartialTxProposal, 'fromPartialTx');
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/atomic-swap/sign-and-push.test.js
```diff
@@ -0,0 +1,127 @@
...
+      .post('/wallet/atomic-swap/tx-proposal/get-my-signatures')
+      .send({ partial_tx: 'partial-tx-data' })
+      .set({ 'x-wallet-id': walletId });
+    TestUtils.logger.debug('[atomic-swap:sign-and-push] should overwrite sigs: get-my-signature', { body: response.body });
+    expect(spy).toHaveBeenCalled();
+
+    const realSigs = response.body.signatures;
+    const parts = realSigs.split('|');
+    parts.splice(-1, 1, '0:cafe');
+    const signatures = [parts.join('|')];
```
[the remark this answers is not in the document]
r4mmer: This was made to test if the signatures on the request body take precedent over the signatures from `signData` but the `signData` call has been removed and this test is now obsolete.

Removed.

__tests__/atomic-swap/tx-proposal-create.test.js
```diff
@@ -0,0 +1,323 @@
...
+    const invalidArray = ['invalid'];
+
+    let response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal')
+      .send({
+        inputs: invalidArray,
+      })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body).toEqual(expect.objectContaining({
```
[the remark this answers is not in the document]
r4mmer: Changed.

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+      .set({ 'x-wallet-id': walletMultisig.walletId });
+    loggers.test.insertLineToLog('atomic-swap[P2SH]: get input data', { body: response.body });
+    expect(response.body).toEqual({
+      success: true,
+      signatures: expect.any(String),
+    });
+
+    const signatureP2SH = response.body.signatures;
+
+    // wallet2: sign-and-push data
```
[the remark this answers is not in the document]
r4mmer: Changed

ATOMIC_SWAP.md
r4mmer: Changed

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+
+    // wallet1: sign data
+    response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal/get-my-signatures')
+      .send({ partial_tx: data })
+      .set({ 'x-wallet-id': wallet1.walletId });
+    loggers.test.insertLineToLog('atomic-swap[2TK P2PKH]: get signatures', { body: response.body });
+    expect(response.body).toEqual({
+      success: true,
+      isComplete: false,
```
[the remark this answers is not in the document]
r4mmer: After some consideration, the previous knowledge of the completeness of signatures is not required for any step, and the only thing is could prevent is the "Transaction is not complete" error message which is not an issue.

I removed the `isComplete` from the get-my-signatures API since it only causes confusion with no actual benefits for the user.

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,259 @@
...
+### POST /wallet/decode
+
+Check what is contained on a proposal string.
+
+```bash
+$ curl -X POST -H "X-Wallet-Id: 123" -H "Content-type: application/json" --data "{\"partial_tx\":\"<proposalString>\"}" http://localhost:8000/wallet/decode
+{
+  "success": true,
+  "tx": {
+    "inputs": [
```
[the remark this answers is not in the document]
r4mmer: Done

.eslintrc
```diff
@@ -34,7 +34,8 @@
     "no-restricted-syntax": ["error", "LabeledStatement", "WithStatement"],
     "no-await-in-loop": "off",
     "no-plusplus": "off",
-    "guard-for-in": "off"
+    "guard-for-in": "off",
+    "no-bitwise": "off",
```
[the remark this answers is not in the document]
r4mmer: never_send_a_human

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,305 @@
...
+
+    // wallet2: sign-and-push data
+    response = await TestUtils.request
+      .post('/wallet/atomic-swap/tx-proposal/sign-and-push')
+      .send({ partial_tx: data, signatures: [signature] })
+      .set({ 'x-wallet-id': wallet2.walletId });
+    loggers.test.insertLineToLog('atomic-swap[HTR P2PKH]: push', { body: response.body });
+    expect(response.body.success).toBe(true);
+
+    await TestUtils.pauseForWsUpdate();
```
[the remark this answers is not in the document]
r4mmer: Done

scripts/helpers/wallet-precalculation.helper.js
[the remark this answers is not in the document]
pedroferreira1: We need to at least explain this in a comment and which tests are better with 1-out-of-1 multisig wallet

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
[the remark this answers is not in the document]
pedroferreira1: Hummm I think the `oneOf` might be good in this case and not in the other. Check the docs: https://express-validator.github.io/docs/check-api.html#oneofvalidationchains-message

> Creates a middleware instance that will ensure at least one of the given chains passes the validation.

It looks like this is what we want here, at least one of them to have length and in the other, we need only one of them. Is that correct?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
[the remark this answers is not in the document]
pedroferreira1: The TODO is in the wrong line

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,335 @@
...
+1. Signing phase
+  - Participants will sign the proposed transaction and exchange the signatures.
+  - A participant will collect the signatures and push the signed transaction on the network.
+
+### Simulation
+
+Let's divide the steps from an atomic-swap using a simulated exchange between Alice and Bob.
+
+1. Alice and Bob agree that Bob will send 10 HTR to Alice in exchange for 20 TKa (Token Alice).
+2. Bob starts a proposal with his side of the swap (sending 10 HTR and receiving 20 TKa to one of his addresses).
```
pedroferreira1, line 25: What are the parameters used here by Bob?

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,335 @@
...
+    - Uses `POST /wallet/atomic-swap/tx-proposal`
+    - He will send the serialized proposal to Alice.
+    - Alice will check the contents of the proposal with the decode api
+3. Alice will update the proposal with her side of the swap (sending 20 TKa and receiving 10 HTR to one of her addresses).
```
pedroferreira1, line 29: Bob already said that he will receive 20 TKa, Alice needs to say the same to the proposal API? Which parameters should be used?

ATOMIC_SWAP.md
```diff
@@ -0,0 +1,335 @@
...
+The process can be divided in 2 phases:
+1. Negotiation phase
+  - Participants will coordinate the content of the atomic-swap.
+1. Signing phase
+  - Participants will sign the proposed transaction and exchange the signatures.
+  - A participant will collect the signatures and push the signed transaction on the network.
+
+### Simulation
+
+Let's divide the steps from an atomic-swap using a simulated exchange between Alice and Bob.
```
pedroferreira1, line 22: I like the example but I feel we should have the requests detailed in all the steps. After this section you start describing all the requests but they could be detailed in each example step.

ATOMIC_SWAP.md
pedroferreira1: I was not exactly talking about the low level difference, it's more about the guide level. If you are explaining to the people who will use the headless wallet, when they should use `receive_tokens` and when they should use `outputs`?

I'm afraid we are creating more parameters and complicating even more the end user decision.

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,477 @@
...
+      .send({
+        partial_tx: data,
+        receive_tokens: [{ value: 140 }],
+        send_tokens: [{ token: tokenTx2.hash, value: 50 }],
+      })
+      .set({ 'x-wallet-id': wallet2.walletId });
+    loggers.test.insertLineToLog('atomic-swap[P2SH]: update proposal 2', { body: response.body });
+    expect(response.body.isComplete).toBe(true);
+
+    data = response.body.data;
```
pedroferreira1, line 374: We should use this data in the decode API to validate that everything is correct, especially all the change outputs. 

If there aren't change outputs, we should change the test to have changes for all wallets and validate each of the change outputs if their addresses belong to their own wallets, so we don't get to any bug similar to the multisig one.

tuliomir (approved):

ATOMIC_SWAP.md
tuliomir: I like the granularity level the current simulation section has. It is long enough to have relevant information but not deep enough to scare developers that are having their first contacts with the feature ( possibly even with the concept of an atomic swap ).

Still, I agree that a list with a more low-level explaining each request could be helpful. For this purpose, what do you think of having an integrated test code embedded on the MD file?
The GitHub syntax for this is just the file URL with the lines specified:
```
https://github.com/HathorNetwork/hathor-wallet-headless/blob/df255e9f0f1126d7488962381a1e5a723026c452/__tests__/integration/atomic-swap.test.js#L56-L125
```

And the result is rendered like the one below:
https://github.com/HathorNetwork/hathor-wallet-headless/blob/df255e9f0f1126d7488962381a1e5a723026c452/__tests__/integration/atomic-swap.test.js#L56-L125

Deeper than that, maybe a sample application on our GitHub with a very commented code connecting to the headless could be a good idea.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -30,86 +32,114 @@ async function buildTxProposal(req, res) {
...
-        txout.index,
-        txout.value,
-        txout.decoded.address,
-        {
-          markAsSelected,
-          token_data: txout.token_data,
-          token: txout.token,
-        },
-      );
+        let heightLocked = false;
```
tuliomir, line 66: I think this validation is important enough to have its own helper method on the lib's `HathorWallet` class itself. A method like `isUtxoUsable` could validate its timelock, reward lock height, _mark as locked_ and others on a single validation response object.

This code extraction doesn't need to be done now, as it's currently only used on this `tx-proposal` method, but it's an important piece of business logic.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -30,86 +32,114 @@ async function buildTxProposal(req, res) {
...
+            throw new Error(`Utxo for transaction ${utxo.txId} and index ${utxo.index} \
+is a block reward and require ${oldWallet.getRewardLockConstant() - blocksMined} more block(s) \
+to be mined before it can be spent.`);
+          }
+        }
+
+        const txout = txData.tx.outputs[utxo.index];
+
+        const addressIndex = req.wallet.getAddressIndex(txout.decoded.address);
+        const addressPath = addressIndex ? `m/44'/${HATHOR_BIP44_CODE}'/0'/0/${addressIndex}` : '';
```
tuliomir, line 80: What about using the account path available on the lib's `constants`? 
```suggestion
        const addressPath = addressIndex ? `${P2PKH_ACCT_PATH}'/0/${addressIndex}` : '';
```

tuliomir (approved):
Approved, I only added some comments that do not affect code logic.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: I agree, maybe we should create an issue for this and include another improvement @pedroferreira1 suggested [here](https://github.com/HathorNetwork/hathor-wallet-lib/pull/408#discussion_r960428742)

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: Changed to use the address path helper from the lib.

__tests__/integration/atomic-swap.test.js
r4mmer: Done.

setupTests-integration.js
```diff
@@ -7,6 +7,28 @@ import {
   precalculationHelpers, WalletPrecalculationHelper,
 } from './scripts/helpers/wallet-precalculation.helper';
 
+expect.extend({
+  toBeInArray(received, expected) {
```
r4mmer, line 11: This custom matcher was added to test that an address is on the known wallet addresses array.
We cannot use the usual `expect(array).toContain(element)` matcher since the place we need this is inside an object matcher (`expect(obj).toMatchObject(...)`)

src/routes/wallet/wallet.routes.js
```diff
@@ -167,20 +170,10 @@ walletRouter.post(
...
-          if (!(/^[0-9a-fA-F]+$/.test(value))) return false;
-          return true;
-        }
-      },
-    },
-  }),
+  oneOf([
+    checkSchema(txHexSchema),
+    checkSchema(partialTxSchema),
+  ], 'Required at least one of txHex or partial-tx'),
```
pedroferreira1, line 176: Isn't it `partial_tx`? And should we change to `partialTx` so we keep the same format we've been using?

src/controllers/wallet/wallet.controller.js
```diff
@@ -259,15 +259,37 @@ async function decodeTx(req, res) {
...
-    const tx = helpersUtils.createTxFromHex(txHex, req.wallet.getNetworkObject());
+    let tx;
+    if (txHex !== null) {
+      tx = helpersUtils.createTxFromHex(txHex, req.wallet.getNetworkObject());
+    } else {
+      const partial = PartialTx.deserialize(partialTx, req.wallet.getNetworkObject());
+      // Validate will check with the fullnode if the partial-tx inputs exists and are valid
+      const valid = await partial.validate();
+      if (!valid) {
+        throw new Error('Partial transaction inconsistent with backend');
```
pedroferreira1, line 282: I don't think we should throw an error here, maybe just 

```
res.status(400).json({
      success: false,
      error: 'Invalid partial transaction hex',
    });
    return;
```

What do you think?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  const proposal = partialTx
+    ? PartialTxProposal.fromPartialTx(partialTx, network) : new PartialTxProposal(network);
+
+  if (sendTokens.utxos && sendTokens.utxos.length > 0) {
+    try {
+      const currentTs = dateFormatter.dateToTimestamp(new Date());
+      for (const utxo of sendTokens.utxos) {
+        const txData = await new Promise((resolve, reject) => {
```
pedroferreira1, line 57: I don't think we should do any validations here. The lib code should do everything needed for the integration, we can't expect every use case that will use the lib to implement all this code.

Apart from that, I don't feel we should have this validation at all, I might be preparing a transaction that will be executed in a few hours but I want to start preparing it, so I can get the other person's signature. If the transaction has any problem, the full node will reject it.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -0,0 +1,318 @@
...
+    const response = await sendTransaction.runFromMining();
+    res.send({ success: true, ...mapTxReturn(response) });
+  } catch (err) {
+    res.send({ success: false, error: err.message });
+  } finally {
+    lock.unlock(lockTypes.SEND_TX);
+  }
+}
+
+async function getInputData(req, res) {
```
pedroferreira1, line 225: When does this method should be used?

r4mmer: This is to extract the input data in the atomic-swap format from a transaction hex.
It should be used to extract these signatures made in other ways (e.g. offline signing) but the main reason this API was implemented was to use with MultiSig wallets.

MultiSig wallets have their own signing APIs and finish the process with a partially signed transaction hex (partially since only the P2SH inputs are signed).
This API is meant to extract these signatures so that MultiSIg wallets can send their signatures to an atomic-swap.

src/routes/wallet/wallet.routes.js
r4mmer: Changed to `partial_tx`.

I kept this argument as snake case because we have other API arguments as snake case (e.g. `change_address`, `max_utxos`, `create_mint`) while `txHex` seems to be an exception.
But I have no problem changing this to `partialTx` since we already use this as the variable name.
What do you think?

pedroferreira1: We can keep `partial_tx`, no problem

src/controllers/wallet/wallet.controller.js
pedroferreira1: No, I think it's good

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
pedroferreira1: I think this should be in the docstring of the API

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,726 @@
...
+      .post('/wallet/atomic-swap/tx-proposal')
+      .send({
+        send: {
+          tokens: [
+            { token: tokenTx1.hash, value: 5 },
+            { token: '00', value: 10 },
+          ],
+        },
+        receive: {
+          tokens: [{ token: tokenTx2.hash, value: 15 }],
```
pedroferreira1, line 397: aI think we should have at least one test choosing the address and validating it was really sent to it.

__tests__/integration/atomic-swap.test.js
pedroferreira1: Where was this done?

r4mmer:
> It's still the same, right?

More or less the same, there were some api changes (names of arguments mostly, e.g. `send_tokens` to `send.tokens`) and some mock strategies changed to reflect the wallet-lib changes.

I changed the location for organization and to facilitate testing this feature.
The idea is to make it easier to find the tests related to this feature and create new ones if needed.

__tests__/integration/atomic-swap.test.js
r4mmer: Created a test for specifying addresses on the atomic-swap (both for send change and receive)

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,726 @@
...
+import { WalletHelper } from './utils/wallet-helper';
+import { singleMultisigWalletData } from '../../scripts/helpers/wallet-precalculation.helper';
+import { loggers } from './utils/logger.util';
+
+describe('send tx (HTR)', () => {
+  let wallet1;
+  let wallet2;
+  let walletMultisig;
+
+  let fundsTx1B;
```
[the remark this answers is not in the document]
r4mmer: Outdated.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: During our talk we decided to remove the fullnode api call here and get the transaction from the wallet history with `getTx`.
The reason for this is that while this removes the possibility to add utxos from another wallet it removes an external dependency (i.e. call to the fullnode) when building the proposal.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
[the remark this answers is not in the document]
r4mmer: The comment has been moved to the correct line since this comment.

__tests__/integration/atomic-swap.test.js
r4mmer: Sorry, this was done for the multiple tokens tests and not for the multisig one.
Added this check for the multisig test.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
r4mmer: Added docstrings

tuliomir (approved):
I only added minor text and test suggestions. One [last comment from a previous review](https://github.com/HathorNetwork/hathor-wallet-headless/pull/211/files#r936082025) was left unanswered, though.

__tests__/integration/atomic-swap.test.js
```diff
@@ -0,0 +1,900 @@
...
+        isChange: false,
+        token: '00',
+        value: 4,
+        authorities: 0,
+        decodedScript: expect.objectContaining({
+          address: expect.objectContaining({ base58: recvAddr }),
+        }),
+      })],
+    });
+  });
```
pedroferreira1, line 899: I would only add one more test here:

1. Create a proposal in a way that a change output must be created
2. Don't select the change address
3. Validate that the automatically selected change address belongs to the wallet correctly

pedroferreira1 (approved):

__tests__/integration/atomic-swap.test.js
r4mmer: Done

Issue #223, opened 2022-07-26
Get configuration string of tokens


pedroferreira1:
# Motivation

Use cases need to share created tokens configuration string with their users, then we must have an easy way to get this information using the headless wallet. Nowadays the use case must go to the explorer to see this data, or create a custom js code.

# Summary

Some options to solve the problem are described below but they must be discussed in more details.

1. Create a new API that receives the token uid and returns the configuration string.
2. In the token creation (and NFT creation) API, return the config string for the created token.

tuliomir:
Closed by #225

PR #225, opened 2022-07-27
feat: add new endpoint to return token configuration string

pedroferreira1:
### Acceptance Criteria

- Add the token configuration string in the response of the endpoints to create token and create NFT.
- Add new endpoint to get a configuration string of a given token from its uid.
- Add tests for the new endpoint and the new response field in old APIs.
- Update API docs.

### Security Checklist

### DIFF
--- a/__tests__/configuration_string.test.js
+++ b/__tests__/configuration_string.test.js
@@ -3,14 +3,14 @@ import TestUtils from './test-utils';
 describe('configuration string', () => {
   it('should return 400 without a token parameter', async () => {
     const response = await TestUtils.request
-      .get('/configuration-string')
+      .get('/configuration-string');
     expect(response.status).toBe(400);
   });
 
   it('should return 200 with token in parameter', async () => {
     const response = await TestUtils.request
       .get('/configuration-string')
-      .query({ token: '1234' })
+      .query({ token: '1234' });
     expect(response.status).toBe(200);
   });
 });
--- a/__tests__/configuration_string.test.js
+++ b/__tests__/configuration_string.test.js
@@ -1,0 +1,16 @@
+import TestUtils from './test-utils';
+
+describe('configuration string', () => {
+  it('should return 400 without a token parameter', async () => {
+    const response = await TestUtils.request
+      .get('/configuration-string')
+    expect(response.status).toBe(400);
+  });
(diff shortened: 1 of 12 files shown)


pedroferreira1:
Closes #223

__tests__/configuration_string.test.js
```diff
@@ -0,0 +1,16 @@
+import TestUtils from './test-utils';
+
+describe('configuration string', () => {
+  it('should return 400 without a token parameter', async () => {
+    const response = await TestUtils.request
+      .get('/configuration-string');
+    expect(response.status).toBe(400);
+  });
+
+  it('should return 200 with token in parameter', async () => {
```
luislhl, line 10: Should we have test cases for when the token is passed but is invalid, and for when it's passed and valid?

In other words, tests for the cases when `success` it True and False in the response.

src/controllers/index.controller.js
```diff
@@ -222,9 +223,35 @@ function multisigPubkey(req, res) {
   });
 }
 
+function getConfigurationString(req, res) {
+  const validationResult = parametersValidation(req);
```
luislhl, line 227: I thought it slightly confusing that the execution of this validation is separated from the definition of what's being validated., without even a reference. I had to dig for a while to find out where we define what to validate (in the `index.routes.js`)

But I believe this is something broader than the scope of this PR, since the other routes are probably using the same pattern.

r4mmer: I also this this is strange but it's how the `express-validator` works.
I don't think this is an issue. It's strange but it's the more generic approach, the developer can choose how this error is treated (return 200 with error, 400, noop, log, error service, etc)

__tests__/configuration_string.test.js
pedroferreira1: Testing this in the mock test is not that good, that's why I try to focus my tests on the integration tests.

The success tests is already implemented in the create token and create nft test files. I've added a new test file to run the invalid token uid test: https://github.com/HathorNetwork/hathor-wallet-headless/pull/225/commits/0b9b19d305314ae0cf9ee8f0185a67ada6e942e1

r4mmer (approved):

PR #229, opened 2022-08-05
fix: set store before creating proposal

r4mmer:
### Security Checklist

### DIFF
--- a/__tests__/integration/multisig.test.js
+++ b/__tests__/integration/multisig.test.js
@@ -269,7 +269,7 @@ describe('send tx (HTR)', () => {
   });
 
   it('should mount the transaction with the correct change outputs', async () => {
-    const network = hathorLib.Network('privatenet');
+    const network = new hathorLib.Network('privatenet');
     const burnAddress = TestUtils.getBurnAddress();
     const response = await TestUtils.request
       .post('/wallet/p2sh/tx-proposal')
--- a/__tests__/integration/multisig.test.js
+++ b/__tests__/integration/multisig.test.js
@@ -267,4 +267,35 @@ describe('send tx (HTR)', () => {
     expect(response.body.success).toBe(false);
     expect(response.body.error).toBeDefined();
   });
+
+  it('should mount the transaction with the correct change outputs', async () => {
+    const network = hathorLib.Network('privatenet');
+    const burnAddress = TestUtils.getBurnAddress();
+    const response = await TestUtils.request
+      .post('/wallet/p2sh/tx-proposal')
+      .send({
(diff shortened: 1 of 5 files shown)


andreabadesso (approved):

src/controllers/wallet/p2sh/tx-proposal.controller.js
```diff
@@ -33,6 +34,7 @@ async function buildTxProposal(req, res) {
     }
   }
   try {
+    storage.setStore(req.wallet.store);
```
pedroferreira1, line 37: I would just add a comment here saying that this should be in the lib and that we should move it there as soon as we have a more complete method to do it in the wallet class.

__tests__/integration/multisig.test.js
```diff
@@ -267,4 +267,35 @@ describe('send tx (HTR)', () => {
...
+      const decoded = output.parseScript(network);
+      if (decoded.address.base58 === burnAddress) {
+        // This is the intended output
+        expect(decoded.getType()).toBe('p2pkh');
+        expect(output.value).toBe(10);
+        continue;
+      }
+
+      // this is a change address and should be p2sh
+      expect(decoded.getType()).toBe('p2sh');
```
pedroferreira1, line 298: Not only p2sh, I would test that it belongs to the wallet sending the transaction as well

__tests__/integration/multisig.test.js
```diff
@@ -267,4 +267,35 @@ describe('send tx (HTR)', () => {
     expect(response.body.success).toBe(false);
     expect(response.body.error).toBeDefined();
   });
+
+  it('should mount the transaction with the correct change outputs', async () => {
```
pedroferreira1, line 271: Do we have a test that we use `tx-proposal` with the `change_address` parameter filled?

__tests__/integration/multisig.test.js
r4mmer: The underlying wallet-lib method already makes this check, should we also make this check?

__tests__/integration/multisig.test.js
r4mmer: Yes, and we test when the change address belongs to the wallet and that it fails when it does not belong to the wallet.

__tests__/integration/multisig.test.js
r4mmer: Done

src/controllers/wallet/p2sh/tx-proposal.controller.js
r4mmer: Done

pedroferreira1 (approved):

PR #240, opened 2022-10-03
feat: readonly wallets

r4mmer:
### Acceptance Criteria
- Implement the [design](https://github.com/HathorNetwork/internal-issues/issues/130)
- Create an error handler for the write protection error.
- Create endpoints for creating and pushing transactions from readonly wallets.

### Security Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1197,9 +1197,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.40.4",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.40.4.tgz",
-      "integrity": "sha512-lGLENhyDEpzI/SF+I1U/oUCd4vjCcVb7B1TtpCXhsLHAmb8vv16IrEclm2JCyH0bXG4kZldkEGLepnS97K0vOw==",
+      "version": "0.41.0",
+      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.41.0.tgz",
+      "integrity": "sha512-YXQeqP6v+hnN1o2iHELSvT2X6GD86lWt6RWmEbw1nazWPB4R88Yx4GMBtf4yH9gr6r0LZ001aOnIF6EA8uM0Ww==",
       "requires": {
         "axios": "^0.21.4",
         "bitcore-lib": "^8.25.10",
@@ -2584,9 +2584,9 @@
       }
     },
     "bitcore-lib": {
-      "version": "8.25.28",
-      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.28.tgz",
-      "integrity": "sha512-UrNHh0Ba8GUiHUYRmm2IKlb8eomsbvk/Z6oQdaOPQoLiamiKnu45pAMqtcHg06wMDF8at54oIdoD2WEU+TQujw==",
+      "version": "8.25.40",
+      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.40.tgz",
+      "integrity": "sha512-mb6kvfhoiIdoyFsDlhIFVst3HpeGjGYBf0XDxTdZ+H07EC4JuiViA3bnQ5uZbZjHFngEl0GTPaoK1Zaolutw4A==",
       "requires": {
         "bech32": "=2.0.0",
         "bip-schnorr": "=0.6.4",
@@ -2606,11 +2606,11 @@
       }
     },
     "bitcore-mnemonic": {
-      "version": "8.25.28",
-      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.28.tgz",
-      "integrity": "sha512-z4EB7r1JYqJCl31rpfm9qZX24tJ+PPvgAk7xvD6x8pqyv60j7hDTn1zt2mDQu0IxGlGMSnQmGmddZf3UcQXQ8w==",
+      "version": "8.25.40",
+      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.40.tgz",
+      "integrity": "sha512-+4tkEwt+F0G1rLGLKCe9m8BhirUu0KNqzvjXvlxmO5S65MelHyAns5/R1CeKH3jEZJDQLghgKPwITfIbe+/uIg==",
       "requires": {
-        "bitcore-lib": "^8.25.28",
+        "bitcore-lib": "^8.25.40",
         "unorm": "^1.4.1"
       }
     },
@@ -4600,9 +4600,9 @@
       "integrity": "sha512-GRnmB5gPyJpAhTQdSZTSp9uaPSvl09KoYcMQtsB9rQoOmzs9dH6ffeccH+Z+cv6P68Hu5bC6JjRh4Ah/mHSNRw=="
     },
     "follow-redirects": {
-      "version": "1.15.1",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.1.tgz",
-      "integrity": "sha512-yLAMQs+k0b2m7cVxpS1VKJVvoz7SS9Td1zss3XRwXj+ZDH00RJgnuLx7E44wx02kQLrdM3aOOy+FpzS7+8OizA=="
+      "version": "1.15.2",
+      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.2.tgz",
+      "integrity": "sha512-VQLG33o04KaQ8uYi2tVNbdrWp1QWxNNea+nmIB4EVM28v0hmP17z7aG1+wAkNzVq4KeXTq3221ye5qTJP91JwA=="
     },
     "for-each": {
       "version": "0.3.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1197,9 +1197,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.41.0",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.41.0.tgz",
-      "integrity": "sha512-YXQeqP6v+hnN1o2iHELSvT2X6GD86lWt6RWmEbw1nazWPB4R88Yx4GMBtf4yH9gr6r0LZ001aOnIF6EA8uM0Ww==",
+      "version": "0.41.1",
+      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.41.1.tgz",
+      "integrity": "sha512-s+bwie+8FLC7cPyX2LtYrCut9QmmRplAiwTxy2+D2VhTbBeydjlI8GWKFgpqm7TOQ2yzRYwz516tGqOCVJjfNg==",
       "requires": {
         "axios": "^0.21.4",
         "bitcore-lib": "^8.25.10",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2584,9 +2584,9 @@
       }
     },
     "bitcore-lib": {
-      "version": "8.25.40",
-      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.40.tgz",
-      "integrity": "sha512-mb6kvfhoiIdoyFsDlhIFVst3HpeGjGYBf0XDxTdZ+H07EC4JuiViA3bnQ5uZbZjHFngEl0GTPaoK1Zaolutw4A==",
+      "version": "8.25.28",
+      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.28.tgz",
+      "integrity": "sha512-UrNHh0Ba8GUiHUYRmm2IKlb8eomsbvk/Z6oQdaOPQoLiamiKnu45pAMqtcHg06wMDF8at54oIdoD2WEU+TQujw==",
       "requires": {
         "bech32": "=2.0.0",
         "bip-schnorr": "=0.6.4",
@@ -2606,11 +2606,11 @@
       }
     },
     "bitcore-mnemonic": {
-      "version": "8.25.40",
-      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.40.tgz",
-      "integrity": "sha512-+4tkEwt+F0G1rLGLKCe9m8BhirUu0KNqzvjXvlxmO5S65MelHyAns5/R1CeKH3jEZJDQLghgKPwITfIbe+/uIg==",
+      "version": "8.25.28",
+      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.28.tgz",
+      "integrity": "sha512-z4EB7r1JYqJCl31rpfm9qZX24tJ+PPvgAk7xvD6x8pqyv60j7hDTn1zt2mDQu0IxGlGMSnQmGmddZf3UcQXQ8w==",
       "requires": {
-        "bitcore-lib": "^8.25.40",
+        "bitcore-lib": "^8.25.28",
         "unorm": "^1.4.1"
       }
     },
@@ -4600,9 +4600,9 @@
       "integrity": "sha512-GRnmB5gPyJpAhTQdSZTSp9uaPSvl09KoYcMQtsB9rQoOmzs9dH6ffeccH+Z+cv6P68Hu5bC6JjRh4Ah/mHSNRw=="
     },
     "follow-redirects": {
-      "version": "1.15.2",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.2.tgz",
-      "integrity": "sha512-VQLG33o04KaQ8uYi2tVNbdrWp1QWxNNea+nmIB4EVM28v0hmP17z7aG1+wAkNzVq4KeXTq3221ye5qTJP91JwA=="
+      "version": "1.15.1",
+      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.1.tgz",
+      "integrity": "sha512-yLAMQs+k0b2m7cVxpS1VKJVvoz7SS9Td1zss3XRwXj+ZDH00RJgnuLx7E44wx02kQLrdM3aOOy+FpzS7+8OizA=="
     },
     "for-each": {
       "version": "0.3.3",
--- a/src/api-docs.js
+++ b/src/api-docs.js
@@ -52,6 +52,10 @@ const apiDoc = {
                     type: 'string',
                     description: '24-words seed separated with spaces. This parameter is incompatible with seedKey. Either seed or seedKey are required.'
                   },
+                  xpubkey: {
+                    type: 'string',
+                    description: 'Account level xpubkey. This will be used to start wallet on readonly mode.',
+                  },
                   multisig: {
                     type: 'boolean',
                     description: 'Start as a multisig wallet. Requires multisig configuration.'
@@ -857,7 +861,7 @@ const apiDoc = {
         },
       },
     },
-    '/wallet/tx-proposal/push-tx': {
+    '/push-tx': {
       post: {
         summary: 'Push a transaction from the txHex.',
         parameters: [
@@ -872,7 +876,7 @@ const apiDoc = {
           },
         ],
         requestBody: {
-          description: 'Transaction hex and signatures',
+          description: 'Signed transaction hex',
           required: true,
           content: {
             'application/json': {
--- a/src/api-docs.js
+++ b/src/api-docs.js
@@ -56,6 +56,10 @@ const apiDoc = {
                     type: 'boolean',
                     description: 'Start as a multisig wallet. Requires multisig configuration.'
                   },
+                  multisigKey: {
+                    type: 'string',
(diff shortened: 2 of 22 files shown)


src/controllers/index.controller.js
```diff
@@ -22,11 +23,11 @@ function docs(req, res) {
 }
 
 function start(req, res) {
-  // We expect the user to send the seed he wants to use
-  if (!('seedKey' in req.body) && !('seed' in req.body)) {
+  // We expect the user to either send the seed of an xpubkey he wants to use.
```
andreabadesso, line 26:
```suggestion
  // We expect the user to either send the seed or an xpubkey he wants to use.
```

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,160 @@
...
+async function buildTxProposal(req, res) {
+  const validationResult = parametersValidation(req);
+  if (!validationResult.success) {
+    res.status(400).json(validationResult);
+    return;
+  }
+
+  // build a transaction hex
+  const { wallet } = req;
+  const network = req.wallet.getNetworkObject();
```
andreabadesso, line 28:
```suggestion
  const network = wallet.getNetworkObject();
```

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,160 @@
...
+  // Expects array of objects with {'txId', 'index'}
+  let inputs = req.body.inputs || [];
+  const changeAddress = req.body.change_address || null;
+
+  if (inputs.length > 0) {
+    if (inputs[0].type === 'query') {
+      const query = inputs[0];
+      inputs = [];
+      // query processing
+
```
andreabadesso, line 58:
```suggestion

      // query processing
```

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,160 @@
...
+      inputs = inputs.map(input => ({ txId: input.hash, index: input.index }));
+    }
+  }
+
+  try {
+    const sendTransaction = new SendTransaction({
+      outputs,
+      inputs,
+      changeAddress,
+      pin: '123',
```
andreabadesso, line 92: What is this? We always use `123` as the PIN here? Should this be a constant?

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+    connectionTimeout: config.connectionTimeout,
+  });
+
+  // Previous versions of the lib would have password and pin default as '123'
+  // We currently need something to be defined, otherwise we get an error when starting the wallet
+  const walletConfig = {
+    xpub,
+    connection,
+    password: '123',
+    pinCode: '123',
```
andreabadesso, line 30: Same as comment below, this seems so weird to me, can't we use an empty string here if it's a required param?

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  // tokenUid is optional but if not passed as parameter the wallet will use HTR
+  if (config.tokenUid) {
+    walletConfig.tokenUid = config.tokenUid;
+  }
+
+  // Wallet addresses pre-calculation, usually for speeding up tests
+  if (preCalculatedAddresses && preCalculatedAddresses.length) {
+    console.log(`Received pre-calculated addresses`, preCalculatedAddresses);
```
andreabadesso, line 41: Remove this log?

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  if (passphrase) {
+    // If config explicitly allows the /start endpoint to have a passphrase
+    const allowPassphrase = config.allowPassphrase || false;
+
+    if (!allowPassphrase) {
+      // To use a passphrase on /start POST request
+      // the configuration of the headless must explicitly allow it
+      console.error('Failed to start wallet because using a passphrase is not allowed by the current config. See allowPassphrase.');
```
andreabadesso, line 95: Why do we need to log this? We are already handling the error properly

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+      throw new WalletStartError('Failed to start wallet. To use a passphrase you must explicitly allow it in the configuration file. Using a passphrase completely changes the addresses of your wallet, only use it if you know what you are doing.');
+    }
+    walletConfig.passphrase = passphrase;
+  }
+
+  // Wallet addresses pre-calculation, usually for speeding up tests
+  if (preCalculatedAddresses && preCalculatedAddresses.length) {
+    console.log(`Received pre-calculated addresses`, preCalculatedAddresses);
```
andreabadesso, line 103: Remove?

andreabadesso (changes requested):

src/helpers/wallet.helper.js
r4mmer: This was on the original start method especially for debug since the precalculated addresses are used on tests.

src/helpers/wallet.helper.js
r4mmer: This is the same as the other log, to debug the received precalculated addresses

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: The wallet headless has always used '123' as pin and password due to a wallet-lib requirement.
We can have a separate issue to make this refactor since this hardcoded pin is used on other places as well

src/controllers/index.controller.js
r4mmer: Fixed

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: Done

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: Tracking this on #246

src/helpers/wallet.helper.js
r4mmer: Tracking this on #246

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+  // tokenUid is optional but if not passed as parameter the wallet will use HTR
+  if (config.tokenUid) {
+    walletConfig.tokenUid = config.tokenUid;
+  }
+
+  // Wallet addresses pre-calculation, usually for speeding up tests
+  if (preCalculatedAddresses && preCalculatedAddresses.length) {
+    console.log(`Received pre-calculated addresses`, preCalculatedAddresses);
+    walletConfig.preCalculatedAddresses = preCalculatedAddresses;
+  }
```
pedroferreira1, line 43: All those lines are the same for both methods. Why don't you keep this code in the `/start` controller and send here after it's handled?

src/helpers/wallet.helper.js
```diff
@@ -0,0 +1,108 @@
...
+import config from '../config';
+
+export class WalletStartError extends Error {}
+
+export function getReadonlyWallet({
+  xpub,
+  multisigData = null,
+  preCalculatedAddresses = [],
+}) {
+  const connection = new Connection({
```
pedroferreira1, line 18: The connection is the same for both methods. Why don't you create before and have it as a parameter?

src/schemas.js
```diff
@@ -171,3 +170,247 @@ export const atomicSwapCreateSchema = {
...
+          if (!(/^[0-9a-fA-F]+$/.test(signature)
+            && walletUtils.isXpubKeyValid(xpub))) return false;
+        }
+        return true;
+      }
+    },
+  },
+  type: {
+    customSanitizer: {
+      options: () => 'p2sh',
```
pedroferreira1, line 413: What's this `p2sh` used for?

src/schemas.js
```diff
@@ -171,3 +170,247 @@ export const atomicSwapCreateSchema = {
...
+      },
+    },
+    toInt: true,
+  },
+  'signatures.*.data': {
+    in: ['body'],
+    errorMessage: 'Invalid signature data',
+    isString: true,
+    custom: {
+      options: value => /^[0-9a-fA-F]+$/.test(value),
```
pedroferreira1, line 354: We have this hex validation test in many places here. Maybe just create a method at the beginning of the file to reuse this. What do you think?

src/schemas.js
```diff
@@ -171,3 +170,247 @@ export const atomicSwapCreateSchema = {
...
+  },
+  change_address: {
+    in: ['body'],
+    errorMessage: 'Invalid change address',
+    isString: true,
+    optional: true,
+  },
+};
+
+export const txInputSchema = {
```
pedroferreira1, line 241: Shouldn't this have the `inputs` array definition in the beginning? Or you don't need that?

```
inputs: {
    in: ['body'],
    errorMessage: 'Invalid inputs array',
    isArray: true,
    notEmpty: true,
  },
```

src/schemas.js
```diff
@@ -171,3 +170,247 @@ export const atomicSwapCreateSchema = {
...
+  'outputs.*.token': {
+    in: ['body'],
+    errorMessage: 'Invalid token uid',
+    isString: true,
+    optional: true,
+    custom: {
+      // Test if token uid is a 64 character hex string or 00
+      options: value => /^(00|[0-9a-fA-F]{64})$/.test(value),
+    },
+  },
```
pedroferreira1, line 208: Should we have `outputs.*.data` here as well? And `outputs.*.type`?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,247 @@
...
+        };
+        const utxos = getUtxosToFillTx(wallet, tokenObj.amount, queryOptions);
+        if (!utxos) {
+          const response = {
+            success: false,
+            error: 'No utxos available for the query filter for this amount.',
+            token: tokenUid
+          };
+          res.send(response);
+          lock.unlock(lockTypes.SEND_TX);
```
pedroferreira1, line 78: Why do you have the `unlock` if you don't have the `lock`?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,247 @@
...
+      fullTxData,
+      '123',
+      { getSignature: false },
+    );
+    const tx = helpersUtils.createTxFromData(preparedData, network);
+    res.send({ success: true, txHex: tx.toHex(), dataToSignHash: tx.getDataToSignHash().toString('hex') });
+  } catch (err) {
+    res.send({ success: false, error: err.message });
+  } finally {
+    lock.unlock(lockTypes.SEND_TX);
```
pedroferreira1, line 112: Why do you have the `unlock` if you don't have the `lock`?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,247 @@
...
+    res.status(400).json(validationResult);
+    return;
+  }
+
+  // build a transaction hex
+  const { wallet } = req;
+  const network = wallet.getNetworkObject();
+  const { outputs } = req.body;
+
+  const tokens = new Map();
```
pedroferreira1, line 35: I feel most of this code is pretty much the same as the `send-tx` API, isn't it? Can we reuse it so we don't duplicate this code?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,247 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const {
+  constants: hathorLibConstants,
+  helpersUtils,
+  SendTransaction,
+  transaction,
+  transaction: transactionUtils,
```
pedroferreira1, line 13: Do you have two `transaction` being imported here?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
```diff
@@ -0,0 +1,247 @@
...
+  if (type === 'p2pkh') {
+    const { signature } = req.body.signature;
+    inputData = transactionUtils.createInputData(
+      Buffer.from(signature, 'hex'),
+      walletUtils.getPublicKeyFromXpub(accessData.xpubkey, index).toBuffer(),
+    );
+  } else if (type === 'p2sh') {
+    // check if the loaded wallet is MultiSig
+    const multisigData = accessData.multisig;
+    if (multisigData) {
```
pedroferreira1, line 204: Shouldn't this be `!multisigData`?

pedroferreira1 (commented):
It's also missing the new APIs in the `api-docs.js` file and the changes in old APIs.

src/controllers/index.controller.js
```diff
@@ -82,29 +58,44 @@ function start(req, res) {
...
+      message: `Failed to start wallet with wallet id ${walletID}`,
+      errorCode: API_ERROR_CODES.WALLET_ALREADY_STARTED,
+    });
+    return;
+  }
+
   let multisigData = null;
   if (('multisig' in req.body) && (req.body.multisig !== false)) {
-    if (!(config.multisig && (seedKey in config.multisig))) {
+    const multisigKey = req.body.seedKey || req.body.multisigKey;
```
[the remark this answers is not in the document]
r4mmer: Because a readonly wallet cannot be initialized with a `seedKey` since there is no seed, and multisig could only be initialized with a `seedKey`.

src/controllers/index.controller.js
```diff
@@ -116,69 +107,65 @@ function start(req, res) {
...
+        res.send({
+          success: false,
+          message: e.message,
+        });
+        return;
+      }
+      // Unhandled error
+      throw e;
+    }
+  } else if (('seed' in req.body) || ('seedKey' in req.body)) {
```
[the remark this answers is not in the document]
r4mmer: Removed the if, but added a comment with the case this `else` refers to.

src/helpers/wallet.helper.js
r4mmer: These are on the wallet-helper because it directly affects the `walletConfig` which is internal to these method.

The `preCalculatedAddresses` are extracted on the `/start` handler here we just add them to the `walletConfig` if they exist.

The only way to move this to the `/start` handler would be to make these methods only build the `walletConfig` and we could update it on the `/start` handler before initiating the wallet.

package-lock.json
```diff
@@ -1197,9 +1197,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.40.4",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.40.4.tgz",
-      "integrity": "sha512-lGLENhyDEpzI/SF+I1U/oUCd4vjCcVb7B1TtpCXhsLHAmb8vv16IrEclm2JCyH0bXG4kZldkEGLepnS97K0vOw==",
+      "version": "0.41.0",
```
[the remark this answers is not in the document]
r4mmer: lodash is not a new dependency, but the linter required that we explicitly add it to dependencies because we import it in the project.

Just updated the package-lock with the version 0.41.1, when i removed the bitcore-lib changes I must've removed this as well.

src/helpers/wallet.helper.js
r4mmer: Done

src/schemas.js
r4mmer: This is not an argument, the sanitizer will add the `type: "p2sh"` or `type: "p2pkh"` depending on which validation chain is resolved for the `oneOf`.

This way we can have different arguments based on the `p2sh` or `p2pkh` and this `type` will indicate for the controller which one it is.

src/schemas.js
r4mmer: We have 2 chains of validation, the `txBuildSchema` and the `oneOf([queryInputSchema, txInputSchema])`.
the `txBuildSchema` already checks for inputs array, if the array has length 1 with `type: "query"` it will be recognized as a query input, otherwise it will be a tx input array, which can be empty (meaning the wallet chooses the inputs).

src/schemas.js
r4mmer: Maybe we could have a validation chain for data outputs vs. funds outputs but using the custom validator seemed better.

We could add this validation chain, what do you think?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: I removed the lock but forgot to remove the finally unlock.

src/schemas.js
r4mmer: The hex validation is slightly different in some cases.
Where the length is unknown we just test for hex values, but for tx_ids (i.e. sha256, 64 hex chars) we test the length.
And if it's a token uid we may accept HTR `00` as well.

If we were to have a specific validation method for token uids and signatures maybe it should come from the wallet-lib.
What do you think?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: Removed.

src/controllers/index.controller.js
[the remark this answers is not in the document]
pedroferreira1: Yes, I agree but I still don't understand why you created the new parameter in the body `multisigKey`?

src/helpers/wallet.helper.js
pedroferreira1: I just don't like the duplicated code. If you move this the same way you moved the `connection` code we would prevent this duplicated code.

src/schemas.js
pedroferreira1: Yeah, makes sense. It's fine

src/schemas.js
pedroferreira1: For me it makes sense because we accept data outputs here, right?

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
pedroferreira1: How was it working with the tests like this? Did you add a new test for this to fail?

r4mmer: When running locally I made a filter to run the tx-proposal tests, but I think when I filtered out the multisig tests it also filtered out the multisig cases of the tx-proposal tests. If you check the fixes when the CI was failing most of them were on p2sh cases.

src/api-docs.js
```diff
@@ -56,6 +56,10 @@ const apiDoc = {
                     type: 'boolean',
                     description: 'Start as a multisig wallet. Requires multisig configuration.'
                   },
+                  multisigKey: {
+                    type: 'string',
+                    description: 'Key of the multisig wallet data in the config. This allow wallets to be started without a seedKey, e.g. with the seed on the parameters or from an xpubkey.',
+                  },
                 }
```
pedroferreira1, line 63: It's missing the new `xpubkey` parameter here, right?

src/api-docs.js
```diff
@@ -591,6 +595,458 @@ const apiDoc = {
...
+                    value: { success: false, message: 'Invalid wallet-id parameter.' }
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    '/wallet/tx-proposal/push-tx': {
```
pedroferreira1, line 860: I was thinking about this API here... semantically we could change this to be only `push-tx`. We don't even need a wallet to push the tx, given that we have the hex, right? So it shouldn't be inside the `tx-proposal` group of APIs and not even the `wallet` group of APIs. What do you think?

pedroferreira1, line 860: I was thinking about this API here... semantically we could change this to be only `push-tx`. We don't even need a wallet to push the tx, given that we have the hex, right? So it shouldn't be inside the `tx-proposal` group of APIs and not even the `wallet` group of APIs. What do you think?

src/schemas.js
pedroferreira1: Hummm ok

src/api-docs.js
r4mmer: Done

src/api-docs.js
r4mmer: Added

src/controllers/wallet/tx-proposal/tx-proposal.controller.js
r4mmer: I gave some though to this, but I still think we should keep the implementation separate.
The main reason would be for future implementations, I think we may add multisig and atomic-swap features to the `tx-proposal` subroute, so later the helper we will create to avoid duplicating code will only be used on the `send-tx` api, making the change seem not that useful.
What do you think?

pedroferreira1: I'm not sure I agree. Even after we consolidate multisig and atomic swap APIs into this one, we will need all those parameters and this same logic for getting the correct utxos to create the proposal, won't we?

r4mmer: It depends, the atomic-swap does not need to select utxos when creating the proposal.

pedroferreira1: Does not need to but it might need. It's the same as the `send-tx` API. The user can manually select the input but we can have a utxo selection, then we would need the same logic in both to support querying by address and the common utxo selection.

pedroferreira1 (approved):

andreabadesso (approved):

PR #275, opened 2023-02-23
feat/Atomic Swap Service create

tuliomir:
### Acceptance Criteria
- The headless wallet must be able to create a new proposal on the [`Atomic Swap Service`](https://github.com/HathorNetwork/hathor-atomic-swap-service)
- This will only happen when the parameters `service.is_new` and `service.password` are added to the `/wallet/atomic-swap/tx-proposal` endpoint
- This feature will be disabled by default via a feature flag on the `src/constants` file

### Security Checklist


package.json
```diff
@@ -12,9 +12,9 @@
     "@babel/node": "^7.13.0",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@babel/preset-env": "^7.13.5",
-    "@hathor/wallet-lib": "^0.42.1",
-    "express-validator": "^6.10.0",
```
tuliomir, line 16: The IDE changed the position of the `express-validator` lib on its dependency auto-sorting.

src/services/atomic-swap.service.js
```diff
@@ -30,6 +31,29 @@ const assembleTransaction = (partialTx, signatures, network) => {
...
+ */
+const serviceCreate = async (partialTx, password) => {
+  if (!partialTx) {
+    throw new Error('Invalid PartialTx');
+  }
+  if (!password || !password.length || password.length < 3) {
+    throw new Error('Password must have at least 3 characters');
+  }
+
+  const { success, id } = await swapService.create();
```
pedroferreira1, line 48: Don't you need to use `partialTx` and `password` to create the swap?

package.json
```diff
@@ -12,9 +12,9 @@
...
+    "@hathor/wallet-lib": "^0.46.0",
```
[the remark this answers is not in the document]
tuliomir: Yes we should, added on [d253b46](https://github.com/HathorNetwork/hathor-wallet-headless/pull/275/commits/d253b46751f6e6c64b6580ed98cbe304e6aa1995)

src/services/atomic-swap.service.js
tuliomir: Fixed this on [d253b46](https://github.com/HathorNetwork/hathor-wallet-headless/pull/275/commits/d253b46751f6e6c64b6580ed98cbe304e6aa1995), along with the necessary url configuration on app init.

I'll add a comment about this as well.

src/app.js
```diff
@@ -8,6 +8,8 @@
 import express from 'express';
 import morgan from 'morgan';
 import { config as hathorLibConfig, wallet as oldWalletUtils } from '@hathor/wallet-lib';
+// TODO: Import from '@hathor/wallet-lib/lib/config' as soon as it's available
+import { SWAP_SERVICE_MAINNET_BASE_URL, SWAP_SERVICE_TESTNET_BASE_URL } from './constants';
```
tuliomir, line 12: The next version of the Wallet Lib will have these constants exported ( PR HathorNetwork/hathor-wallet-lib#471 ), so we'll fetch if from there instead duplicating them here.

For now, this is following the pattern used by the Wallet Service of having the URLs on the client application.
https://github.com/HathorNetwork/hathor-wallet/blob/18c8314ea3df6b071d93277a2a797b52948ab455/src/constants.js#L237-L238

pedroferreira1 (commented):
We must have a feature flag to disable the new atomic swap endpoints (or just the part the needs the atomic swap service) until the service is fully working. Otherwise, we would be breaking the endpoints with this PR.

r4mmer (approved):

pedroferreira1 (approved):

PR #277, opened 2023-03-13
feat: Swap service `get`

tuliomir:
### Acceptance Criteria
- The headless wallet must be able to retrieve an existing proposal from the [`Atomic Swap Service`](https://github.com/HathorNetwork/hathor-atomic-swap-service)
- This will happen through a new `/wallet/atomic-swap/tx-proposal/fetch-from-service` endpoint
- This feature will be disabled by default via a feature flag on the `src/constants` file, and requests to it while disabled will return a `405 Method not allowed` error

### Note
No integration tests will be added at this stage, as there is no readily available image of the Atomic Swap Service to be added to the `docker-compose.yml` structure necessary for them.
A partial implementation of these tests, however, is available on branch [`feat/swap-service-integration`](https://github.com/HathorNetwork/hathor-wallet-headless/tree/feat/swap-service-integration) ( [diff with `dev`](https://github.com/HathorNetwork/hathor-wallet-headless/compare/dev...feat/swap-service-integration?expand=1) ) and can be executed on a local environment by running the Atomic Swap Service directly from the source code.

### Security Checklist


__tests__/atomic-swap/tx-proposal-create.test.js
```diff
@@ -265,6 +265,8 @@ describe('create tx-proposal api', () => {
     config.setSwapServiceBaseUrl('http://fake-swap-service');
 
     it('should not interact with the service if the feature flag is disabled', async () => {
+      // Deactivating the global feature flag. All tests from now on will have it available.
+      global.constants.SWAP_SERVICE_FEATURE_TOGGLE = false;
```
tuliomir, line 269: Added this line to the previous PR test to ensure the constants will be always at the expected state.

src/routes/wallet/atomic-swap/tx-proposal.routes.js
```diff
@@ -42,6 +44,12 @@ txProposalRouter.get(
   getLockedUTXOs,
 );
 
+txProposalRouter.post(
+  '/fetch-from-service',
```
pedroferreira1, line 48: I don't think I like the name of this API. Only `fetch` would be enough for me, what do you think?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -156,6 +157,33 @@ async function buildTxProposal(req, res) {
...
+  }
+  const { password, proposal_id: proposalId } = req.body;
+  // A service mediated request must have the identifier and the password as parameters
+  if (!password || !proposalId) {
+    res.status(400).json({
+      success: false,
+      error: 'Parameters proposal_id and password are mandatory when requesting from service'
+    });
+    return;
+  }
```
pedroferreira1, line 176: Don't we validate that with the schema in the route?

__tests__/atomic-swap/tx-proposal-create.test.js
pedroferreira1: I didn't understand this

pedroferreira1 (commented):
It's missing the `api-docs` update with the new API. (I forgot about this in the previous create PR as well, so you should update it also)

__tests__/atomic-swap/tx-proposal-create.test.js
tuliomir: The comment was badly written, indeed. It deactivates this feature flag for this tests only.
Reworded it on [19bf791](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/19bf7914c7512d2e59cdf47ef53a93b90c2eecf9)

src/routes/wallet/atomic-swap/tx-proposal.routes.js
tuliomir: I think it's a good name. Refactored it on [4bdc332](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/4bdc3327fa7aaafd1c4d78bae9b058de7d775068)

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
tuliomir: Yes, we do! Removed the redundant validation on [a8d69f0](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/a8d69f00ae77210cdc91ae491d82dcaf861a9e8a)

src/routes/wallet/atomic-swap/tx-proposal.routes.js
```diff
@@ -42,6 +44,12 @@ txProposalRouter.get(
   getLockedUTXOs,
 );
 
+txProposalRouter.post(
```
pedroferreira1, line 47: Why is it a POST and not a GET?

r4mmer (approved):

src/routes/wallet/atomic-swap/tx-proposal.routes.js
tuliomir: There will be a refactor based on #279 for this route

tuliomir: This route was changed to a `GET` according to design HathorNetwork/rfcs#50 .

r4mmer (approved):

src/services/atomic-swap.service.js
```diff
@@ -129,6 +129,43 @@ const removeAllWalletProposals = async walletId => {
...
+ * @param {string} password Password, length 3 or more
+ * @returns {Promise<AtomicSwapProposal>} Returns the proposal identifier created by the service
+ */
+const serviceGet = async (proposalId, password) => {
+  if (!proposalId) {
+    throw new Error('Invalid proposalId');
+  }
+  if (!password || !password.length || password.length < 3) {
+    throw new Error('Password must have at least 3 characters');
+  }
```
pedroferreira1, line 159: The service should reject creating a proposal with a weak password, right? Maybe it's not necessary to validate this here, the get will just return an error in case it's invalid. Can you identify the `get` error when it's an invalid ID, or invalid password?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -148,6 +149,46 @@ async function buildTxProposal(req, res) {
...
+
+  const requestedProposal = listenedProposals.get(proposalId);
+  const { password } = requestedProposal;
+
+  // Fetching proposal from the service
+  try {
+    const serviceProposal = await atomicSwapService.serviceGet(proposalId, password);
+    res.json({ success: true, proposal: serviceProposal });
+  } catch (err) {
+    // TODO: Verify the status code instead of the message
```
pedroferreira1, line 181: TODO is already done, right?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -335,6 +376,39 @@ async function listenedProposalList(req, res) {
...
+ */
+async function registerProposal(req, res) {
+  const { walletId } = req;
+  const { proposalId } = req.params;
+  const { password } = req.body;
+
+  // Avoid making requests to the service if the proposal is already on local storage
+  const proposalMap = await atomicSwapService.getListenedProposals(walletId);
+  if (proposalMap.has(req.params.proposalId)) {
+    res.send({ success: true });
```
pedroferreira1, line 390: Should we validate if the `password` is the same as the stored one? Maybe return success false in case it's different. "You already have a proposal with this ID registered but with a different password."

src/api-docs.js
```diff
@@ -1625,6 +1678,95 @@ const apiDoc = {
         },
       },
     },
+    '/wallet/atomic-swap/tx-proposal/fetch': {
+      post: {
```
pedroferreira1, line 1682: Isn't it a get?

src/api-docs.js
```diff
@@ -1625,6 +1678,95 @@ const apiDoc = {
...
+            'application/json': {
+              schema: {
+                type: 'object',
+                required: ['proposal_id', 'password'],
+                properties: {
+                  proposal_id: {
+                    description: 'Proposal identifier on the Atomic Swap Service.',
+                    type: 'string',
+                  },
+                  password: {
```
pedroferreira1, line 1708: The fetch API does not have password

src/api-docs.js
```diff
@@ -1625,6 +1678,95 @@ const apiDoc = {
...
+        requestBody: {
+          description: 'Get the requested proposal by identifier from the atomic swap service.',
+          required: true,
+          content: {
+            'application/json': {
+              schema: {
+                type: 'object',
+                required: ['proposal_id', 'password'],
+                properties: {
+                  proposal_id: {
```
pedroferreira1, line 1704: You don't have this parameter in the body

src/api-docs.js
```diff
@@ -1625,6 +1678,95 @@ const apiDoc = {
...
+                      },
+                    }
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
```
pedroferreira1, line 1769: Missing the register API

src/services/atomic-swap.service.js
tuliomir: This was an attempt to reduce the amount of invalid calls the backend could receive, but not really necessary.
Refactored it on [ab27216](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/ab2721629e4cc87ab47e71c4e0de5b86a480d1b3) to a simple "mandatory parameter check" in this case.

As for fully identifying valid ids and passwords, we could use _regexes_ to check for them, but I think it's too much complexity for almost zero benefit in this case: just avoiding one invalid `get` request to the service.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
tuliomir: You're right! Removed the obsolete comment..

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
tuliomir: Good idea. Implemented it and refactored the tests on [b123fad](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/b123fad0010767cbf318d6f310bf621a6afd758e).

src/api-docs.js
tuliomir: Yes, it is! This apiDoc wasn't updated after the last review. Fixed it on [595c9a5](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/595c9a5ffe0cc04e502d7ff1777cd47ba51a4140).

src/api-docs.js
tuliomir: Fixed it on [595c9a5](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/595c9a5ffe0cc04e502d7ff1777cd47ba51a4140).

src/api-docs.js
tuliomir: Fixed it on [595c9a5](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/595c9a5ffe0cc04e502d7ff1777cd47ba51a4140).

src/api-docs.js
tuliomir: Fixed it on [595c9a5](https://github.com/HathorNetwork/hathor-wallet-headless/pull/277/commits/595c9a5ffe0cc04e502d7ff1777cd47ba51a4140).

pedroferreira1 (approved):

PR #278, opened 2023-03-17
feat: new storage

r4mmer:
### Acceptance Criteria
- refactor the wallet to use the new storage scheme
- bump wallet-lib to version with the new storage scheme

### Security Checklist

### DIFF
--- a/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
+++ b/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
@@ -11,9 +11,8 @@ const {
   helpersUtils,
   PartialTxInputData,
   PartialTx,
-  storage,
+  transactionUtils,
   constants: { HATHOR_TOKEN_CONFIG, TOKEN_MINT_MASK, TOKEN_MELT_MASK },
-  wallet: oldWallet,
 } = require('@hathor/wallet-lib');
 const {
   assembleTransaction,
@@ -35,8 +34,6 @@ async function buildTxProposal(req, res) {
     return;
   }
 
-  const network = req.wallet.getNetworkObject();
-
   const sendTokens = req.body.send || { tokens: [] };
   const receiveTokens = req.body.receive || { tokens: [] };
   const changeAddress = req.body.change_address || null;
@@ -56,38 +53,30 @@ async function buildTxProposal(req, res) {
   }
 
   const proposal = partialTx
-    ? PartialTxProposal.fromPartialTx(partialTx, network) : new PartialTxProposal(network);
+    ? PartialTxProposal.fromPartialTx(partialTx, req.wallet.storage) : new PartialTxProposal(req.wallet.storage);
 
   if (sendTokens.utxos && sendTokens.utxos.length > 0) {
     try {
       for (const utxo of sendTokens.utxos) {
-        const txData = req.wallet.getTx(utxo.txId);
+        const txData = await req.wallet.getTx(utxo.txId);
         if (!txData) {
           // utxo not in history
           continue;
         }
         const txout = txData.outputs[utxo.index];
-        if (!oldWallet.canUseUnspentTx(txout, txData.height)) {
+        if (transactionUtils.canUseUtxo({txId: utxo.txId, index: utxo.index}, req.wallet.storage)) {
           // Cannot use this utxo
           continue;
         }
 
-        const addressIndex = req.wallet.getAddressIndex(txout.decoded.address);
-        const addressPath = addressIndex ? req.wallet.getAddressPathForIndex(addressIndex) : '';
+        const addressIndex = await req.wallet.getAddressIndex(txout.decoded.address);
+        const addressPath = addressIndex ? await req.wallet.getAddressPathForIndex(addressIndex) : '';
         let authorities = 0;
-        if (oldWallet.isMintOutput(txout)) {
+        if (transactionUtils.isMint(txout)) {
           authorities += TOKEN_MINT_MASK;
         }
-        if (oldWallet.isMeltOutput(txout)) {
+        if (transactionUtils.isMelt(txout)) {
           authorities += TOKEN_MELT_MASK;
-        }
-
-        let tokenId = txout.token;
-        if (!tokenId) {
-          const tokenIndex = oldWallet.getTokenIndex(txout.token_data) - 1;
-          tokenId = txout.token_data === 0
-            ? HATHOR_TOKEN_CONFIG.uid
-            : txData.tx.tokens[tokenIndex].uid;
         }
 
         utxos.push({
@@ -96,7 +85,7 @@ async function buildTxProposal(req, res) {
           value: txout.value,
           address: txout.decoded.address,
           timelock: txout.decoded.timelock,
-          tokenId,
+          tokenId: txout.token,
           authorities,
           addressPath,
           heightlock: null,
@@ -118,15 +107,14 @@ async function buildTxProposal(req, res) {
   try {
     for (const send of sendTokens.tokens) {
       const token = send.token || HATHOR_TOKEN_CONFIG.uid;
-      proposal.addSend(req.wallet, token, send.value, { utxos, changeAddress, markAsSelected });
+      await proposal.addSend(token, send.value, { utxos, changeAddress, markAsSelected });
     }
 
     for (const receive of receiveTokens.tokens) {
       const token = receive.token || HATHOR_TOKEN_CONFIG.uid;
       const timelock = receive.timelock || null;
       const address = receive.address || null;
-      proposal.addReceive(
-        req.wallet,
+      await proposal.addReceive(
         token,
         receive.value,
         { timelock, address }
@@ -167,13 +155,10 @@ async function getMySignatures(req, res) {
     return;
   }
 
-  const network = req.wallet.getNetworkObject();
   const partialTx = req.body.partial_tx;
 
-  // TODO remove this when we create a method to sign inputs in the wallet facade
-  storage.setStore(req.wallet.store);
   try {
-    const proposal = PartialTxProposal.fromPartialTx(partialTx, network);
+    const proposal = PartialTxProposal.fromPartialTx(partialTx, req.wallet.storage);
     await proposal.signData('123');
     res.send({
       success: true,
@@ -194,15 +179,11 @@ async function signTx(req, res) {
     return;
   }
 
-  const network = req.wallet.getNetworkObject();
-
   const partialTx = req.body.partial_tx;
   const signatures = req.body.signatures || [];
 
-  // TODO remove this when we create a method to sign inputs in the wallet facade
-  storage.setStore(req.wallet.store);
   try {
-    const tx = assembleTransaction(partialTx, signatures, network);
+    const tx = assembleTransaction(partialTx, signatures, req.wallet.storage);
 
     res.send({ success: true, txHex: tx.toHex() });
   } catch (err) {
@@ -226,17 +207,13 @@ async function signAndPush(req, res) {
     return;
   }
 
-  const network = req.wallet.getNetworkObject();
-
   const partialTx = req.body.partial_tx;
   const sigs = req.body.signatures || [];
 
-  // TODO remove this when we create a method to sign inputs in the wallet facade
-  storage.setStore(req.wallet.store);
   try {
-    const transaction = assembleTransaction(partialTx, sigs, network);
+    const transaction = assembleTransaction(partialTx, sigs, req.wallet.storage);
 
-    const sendTransaction = new SendTransaction({ transaction, network });
+    const sendTransaction = new SendTransaction({ transaction, storage: req.wallet.storage });
     const response = await sendTransaction.runFromMining();
     res.send({ success: true, ...mapTxReturn(response) });
   } catch (err) {
@@ -288,21 +265,19 @@ async function getInputData(req, res) {
  */
 async function getLockedUTXOs(req, res) {
   try {
+    const txMap = new Map();
+    for await (const utxo of req.wallet.storage.utxoSelectedAsInputIter()) {
+      if (!txMap.has(utxo.txId)) {
+        txMap.set(utxo.txId, new Set());
+      }
+      txMap.get(utxo.txId).add(utxo.index);
+    }
     const utxos = [];
-    const historyTransactions = req.wallet.getFullHistory();
-    for (const tx of Object.values(historyTransactions)) {
-      const marked = [];
-      for (const [index, output] of tx.outputs.entries()) {
-        if ((!output.spent_by) && output.selected_as_input === true) {
-          marked.push(index);
-        }
-      }
-      if (marked.length !== 0) {
-        utxos.push({
-          tx_id: tx.tx_id,
-          outputs: marked,
-        });
-      }
+    for (const txEntry of txMap.entries()) {
+      utxos.push({
+        txId: txEntry[0],
+        outputs: Array.from(txEntry[1]),
+      });
     }
 
     res.send({ success: true, locked_utxos: utxos });
@@ -334,7 +309,7 @@ async function unlockInputs(req, res) {
     const tx = partial.getTx();
 
     for (const input of tx.inputs) {
-      req.wallet.markUtxoSelected(input.hash, input.index, false);
+      await req.wallet.markUtxoSelected(input.hash, input.index, false);
     }
 
     res.send({ success: true });
--- a/src/controllers/wallet/p2sh/tx-proposal.controller.js
+++ b/src/controllers/wallet/p2sh/tx-proposal.controller.js
@@ -9,7 +9,6 @@ const {
   constants: hathorLibConstants,
   SendTransaction,
   helpersUtils,
-  storage,
 } = require('@hathor/wallet-lib');
 const { parametersValidation } = require('../../../helpers/validations.helper');
 const { lock, lockTypes } = require('../../../lock');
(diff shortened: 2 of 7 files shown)


src/controllers/wallet/p2sh/tx-proposal.controller.js
```diff
@@ -28,22 +27,30 @@ async function buildTxProposal(req, res) {
...
+      inputs,
+      changeAddress,
+    });
+    const txData = await sendTransaction.prepareTxData();
+    txData.version = 1;
+    const tx = transactionUtils.createTransactionFromData(txData, network);
 
     res.send({ success: true, txHex: tx.toHex() });
   } catch (err) {
+    console.error(err);
```
tuliomir, line 53: Did you check the results from this? I usually get a `[object Object]` from this kind of log.

A `console.error(err.stack)` or `err.message` would probably be more helpful

src/controllers/wallet/wallet.controller.js
```diff
@@ -17,19 +17,29 @@ const { mapTxReturn, prepareTxFunds } = require('../../helpers/tx.helper');
...
     statusCode: wallet.state,
     statusMessage: friendlyWalletState[wallet.state],
     network: wallet.getNetwork(),
     serverUrl: wallet.getServerUrl(),
-    serverInfo: wallet.serverInfo,
+    serverInfo: wallet.storage.version,
   };
-  if (wallet.multisig) {
+  /**
+   * While this is valid since we the multisigData is only set when the wallet is multisig
```
tuliomir, line 33:
```suggestion
   * While this is valid since the multisigData is only set when the wallet is multisig
```

src/controllers/wallet/wallet.controller.js
```diff
@@ -17,19 +17,29 @@ const { mapTxReturn, prepareTxFunds } = require('../../helpers/tx.helper');
...
+   * We should actually check if a wallet is multisig from wallet.storage.getWalletType()
+   * But the WalleType enum is not exposed in wallet-lib yet.
```
tuliomir, line 35:
```suggestion
   * But the WalletType enum is not exposed in wallet-lib yet.
```

__tests__/simple-send-tx.test.js
```diff
@@ -137,17 +137,26 @@ describe('simple-send-tx api', () => {
   });
 
   it('should accept a custom token transaction (token as string)', async () => {
+    // Since the loaded wallet is the same in all tests
```
pedroferreira1, line 140: Why was it working before?

__tests__/integration/atomic-swap.test.js
```diff
@@ -106,9 +110,12 @@ describe('send tx (HTR)', () => {
       .get('/wallet/atomic-swap/tx-proposal/get-locked-utxos')
       .set({ 'x-wallet-id': wallet1.walletId });
     loggers.test.insertLineToLog('atomic-swap[lock]: check with lock', { body: response.body });
+    // XXX: We have an extra utxo which should not be returned since its spent.
     expect(response.body).toEqual({
       success: true,
-      locked_utxos: [],
+      locked_utxos: expect.not.arrayContaining([
```
pedroferreira1, line 116: I think these changes are not supposed to happen, we should fix them and keep the old code here.

__tests__/integration/configuration/test-constants.js
```diff
@@ -73,7 +73,7 @@ export const TOKEN_DATA = {
    * @param {number} tokenData "token_data" property from an output
    * @returns {boolean} True if this is an authority output
    */
-  isAuthorityToken: tokenData => transactionUtils.isTokenDataAuthority(tokenData)
+  isAuthorityToken: tokenData => (libConstants.TOKEN_AUTHORITY_MASK & tokenData) > 0,
```
pedroferreira1, line 76: We don't have a method like this in the lib anymore? Maybe it's worth having it

__tests__/integration/melt-tokens.test.js
```diff
@@ -209,18 +209,18 @@ describe('melt tokens', () => {
       .send({
         token: tokenA.uid,
         amount: 100,
-        deposit_address: await wallet1.getAddressAt(5),
+        deposit_address: await wallet1.getAddressAt(6),
```
pedroferreira1, line 212: Why did you change it?

__tests__/integration/send-tx.test.js
```diff
@@ -1229,14 +1230,19 @@ describe('send tx (custom tokens)', () => {
...
+    await TestUtils.pauseForWsUpdate();
+    // One UXTO with 1000 TKB, and another with 990 HTR
     const tx = await wallet3.sendTx({
       fullObject: {
         inputs: [
           {
             hash: tkbTx1.hash,
             index: outputIndexTKB, // Input for token B
           },
+          {
```
pedroferreira1, line 1242: Why did you add one more input?

__tests__/integration/send-tx.test.js
```diff
@@ -1301,26 +1307,6 @@ describe('send tx (custom tokens)', () => {
     expect(changeHtr.total_amount_available).toBe(txOutputSummary.htr.change);
     expect(changeTkb.total_amount_available).toBe(txOutputSummary.tkb.change);
 
-    // Validating that the change addresses are not the same
```
pedroferreira1, line 1304: Why did you remove this validation?

__tests__/integration/send-tx.test.js
```diff
@@ -1495,11 +1485,11 @@ describe('filter query + custom tokens', () => {
 
     // Sending all of them to address 4
     const tx = await wallet1.sendTx({
-      inputs: [{ type: 'query', filter_address: await wallet1.getAddressAt(3) }],
```
pedroferreira1, line 1498: Why did you change it?

__tests__/integration/tx-history.test.js
```diff
@@ -76,12 +76,12 @@ describe('tx-history routes', () => {
       expect(transaction.outputs[1].value).toEqual(fundTx.outputs[1].value);
     }
 
-    // Validating the transactions' order: chronological ASCENDING
```
pedroferreira1, line 79: Was it a bug before, or the behavior changed?

__tests__/integration/utxo-consolidation.test.js
```diff
@@ -288,17 +289,17 @@ describe('utxo-consolidation routes', () => {
       .set({ 'x-wallet-id': wallet1.walletId });
     const consolidateTx = utxoResponse.body;
 
-    // Evaluating consolidation results. Only source combination possible must be [10, 20, 30].
-    expect(consolidateTx.total_amount).toBe(60);
```
pedroferreira1, line 292: Why did the amount changed?

__tests__/integration/utxo-filter.test.js
```diff
@@ -271,28 +274,27 @@ describe('utxo-filter routes', () => {
...
-    const utxo0 = utxosObj.utxos[0];
-    expect(utxo0.address).toBe(transactions.tx10.address);
-    expect(utxo0.amount).toBe(10);
-    expect(utxo0.locked).toBe(false);
-
-    const utxo1 = utxosObj.utxos[1];
-    expect(utxo1.address).toBe(transactions.tx20.address);
-    expect(utxo1.amount).toBe(20);
-    expect(utxo1.locked).toBe(false);
+    // for (const utxoIndex in utxosObj.utxos) {
```
pedroferreira1, line 284: Why did you remove this validation?

__tests__/integration/atomic-swap.test.js
```diff
@@ -86,11 +88,13 @@ describe('send tx (HTR)', () => {
       .get('/wallet/atomic-swap/tx-proposal/get-locked-utxos')
       .set({ 'x-wallet-id': wallet1.walletId });
     loggers.test.insertLineToLog('atomic-swap[lock]: check with lock', { body: response.body });
+    // XXX: We have an extra utxo which should not be returned since its spent.
+    // This is not concerning but it means the selected utxos are not being correctly cleaned
```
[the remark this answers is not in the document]
r4mmer: There is already a [PR](https://github.com/HathorNetwork/hathor-wallet-lib/pull/498) for this.
The reason its not concerning is that the utxos are not being selected, this is the list of utxos that cannot be selected.
The extra utxo here cannot be selected because it is spent, so it should not be here, but it will never be selected as an input either way.

__tests__/integration/configuration/test-constants.js
r4mmer: We already have a method for this. But I think the best solution would be to remove the `isAuthorityToken` altogether since we can use the lib method on the tests instead. What do you think?

__tests__/integration/create-token.test.js
```diff
@@ -121,7 +121,7 @@ describe('create token', () => {
 
     expect(response.status).toBe(200);
     expect(response.body.success).toBe(false);
-    expect(response.body.error).toContain('Change address');
```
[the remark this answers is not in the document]
r4mmer: The wallet was returning the lib error message, since the lib changed the validation error the message changed.
Do you think we should track the messages returned to the user on the lib or the headless?

scripts/helpers/wallet-precalculation.helper.js
```diff
@@ -125,16 +125,11 @@ export class WalletPrecalculationHelper {
         networkName: config.network,
         accountDerivationIndex,
       });
-      const addresses = walletUtils.getAddresses(
```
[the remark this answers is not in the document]
r4mmer: We have `loadAddresses` to derive a range of addresses from the storage instance, it determines if the addresses should be P2PKH or Multisig and we also have methods to generate addresses be their index (e.g. `deriveAddressFromXPubP2PKH` and `deriveAddressFromDataP2SH`).

Since the wallet precalculation helper only generates P2PKH wallets it makes sense to have a method to generate a range of P2PKH addresses but what do you think of having the same loop here and use `deriveAddressFromXPubP2PKH` instead?

__tests__/simple-send-tx.test.js
r4mmer: Because we used to mark the `selected_as_input` on the history object, which gets "reloaded" from the fixture without the `selected_as_input` or the "spent_by" from the previous test.
Since we removed the `selected_as_input` to a separate collection it does not get "auto cleaned".

I used the new wallet but i can change to just un`select_as_input` on the previous test, what do you think?

__tests__/integration/melt-tokens.test.js
r4mmer: I was tracking the changes that made this happen, the key change was that on the old lib we checked the address index with [`maxIndex > lastUsedIndex`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/793bef622269e4462f62f85a35a7046a2b24b282/src/wallet.js#L2347-L2348) so if the max index of the last tx spent is the same as the last used address, the "current address" will not be updated and will point to an used address.

The new lib checks [`lastUsedAddressIndex <= maxIndexUsed`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/aba5242e7bb137a390e97afc601e08ca815b3e39/src/utils/storage.ts#L264-L265) which will update the "current address" so it is always unused.

This difference is why here we choose the change address as the address 4 (which is used) and the current lib chooses the address 5 which is unused. Since the address 5 has another change utxo the `expects` below will fail, that's why I bumped the addresses, but we can also explicitly use the address 4 as change to the same effect, what do you think?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -53,47 +50,43 @@ async function buildTxProposal(req, res) {
...
         }
 
-        let tokenId = txout.token;
-        if (!tokenId) {
-          const tokenIndex = oldWallet.getTokenIndex(txout.token_data) - 1;
-          tokenId = txout.token_data === 0
-            ? HATHOR_TOKEN_CONFIG.uid
-            : txData.tx.tokens[tokenIndex].uid;
-        }
-
```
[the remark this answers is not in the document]
r4mmer: Actually `txout.token` is created as a [`tx.to_json_extended`](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/wallet/resources/thin_wallet/address_history.py#L223) which always adds the token, even for HTR.
What it does not add is a tokens array with the token config, so if was never being used and if it somehow entered it would throw an error.

__tests__/integration/melt-tokens.test.js
r4mmer: This test was ensuring the wrong behavior since the expected change address was an address with a transaction already.
This is explained in the comment at the start of the test

__tests__/integration/send-tx.test.js
r4mmer: The wallet-lib does not mark the address as used when it is chosen as change, so they will not be different.

__tests__/integration/tx-history.test.js
r4mmer: The behavior changed to iterate on the history on timestamp descending.
The previous behavior was loosely iterating on insertion order which was timestamp ascending on this test.
The previous order was not guaranteed but we currently enforce the chronological descending order

__tests__/integration/create-token.test.js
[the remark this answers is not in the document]
tuliomir: I think this deserves to be tracked in both places:
- On the lib, since it serves many of our products and partners;
- On the headless, because this is the final result for the end user

It's a code duplication, but when we are asserting the results it feels more like a redundancy than a code smell. Do you agree?

( by the way, this code is already OK again )

__tests__/address.test.js
```diff
@@ -52,7 +51,8 @@ describe('address api', () => {
   });
 
   it('should return a new address with mark_as_used until the gapLimit is reached', async () => {
-    const gapLimit = walletUtils.getGapLimit();
+    global.config.gapLimit = 20;
+    const gapLimit = 20;
```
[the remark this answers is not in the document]
r4mmer: Done

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -53,47 +50,43 @@ async function buildTxProposal(req, res) {
   }
 
   const proposal = partialTx
-    ? PartialTxProposal.fromPartialTx(partialTx, network) : new PartialTxProposal(network);
+    ? PartialTxProposal.fromPartialTx(partialTx, req.wallet.storage)
```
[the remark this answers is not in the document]
r4mmer: Done

src/controllers/wallet/p2sh/tx-proposal.controller.js
```diff
@@ -28,22 +27,30 @@ async function buildTxProposal(req, res) {
...
-      network
-    );
+    const sendTransaction = new SendTransaction({
+      storage: req.wallet.storage,
+      outputs,
+      inputs,
+      changeAddress,
+    });
+    const txData = await sendTransaction.prepareTxData();
+    txData.version = 1;
```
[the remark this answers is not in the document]
r4mmer: Done

src/helpers/notification.helper.js
```diff
@@ -30,16 +29,19 @@ const {
 /**
  * Calculate the wallet balance for a tx
  *
+ * XXX: This can be calculated using the lib methods
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/integration/utxo-filter.test.js
r4mmer: This has since changed, but we cannot guarantee the order of the utxos now.
The main reason is that since we are getting the most recent utxos first, the first 2 available utxos on the filter would be the one from tx50 and the change from it, but we shuffle when we have a change so its position will not be fixed.

src/controllers/wallet/wallet.controller.js
```diff
@@ -398,6 +436,9 @@ async function createToken(req, res) {
   const address = req.body.address || null;
   const changeAddress = req.body.change_address || null;
   try {
+    if (changeAddress && !await wallet.isAddressMine(changeAddress)) {
```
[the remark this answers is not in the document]
r4mmer: Actually this was not consistent even on the previous lib version.
The wallet-lib would not enforce the change address must be from the wallet depending on the method.
There were even some skipped tests with this observation.

I also believe the change address should be enforced on the lib, and i'm making a PR to do this, but I also think we can check on the headless as well.
What do you think?

__tests__/integration/utxo-filter.test.js
```diff
@@ -284,15 +286,18 @@ describe('utxo-filter routes', () => {
...
+        address: addr0,
+        amount: 850,
+        locked: false,
+      }),
+      expect.objectContaining({
+        address: transactions.tx50.address,
+        amount: 50,
+        locked: false,
+      }),
+    ]));
```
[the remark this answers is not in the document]
r4mmer: Due to the order in which we select the utxos the most recent will come first which means that the tx50 (last tx) outputs will be the first 2 available utxos and they are a 50 tokenA and the change of 850 tokenA.

The previous behavior was to iterate on "insertion order" on the history object, this would loosely follow the order on the tests but was not guaranteed on all cases.

__tests__/integration/utxo-filter.test.js
```diff
@@ -271,7 +274,6 @@ describe('utxo-filter routes', () => {
 
     // Validating filtered response
     expect(utxosObj.utxos).toHaveProperty('length', 2);
-    expect(utxosObj.total_amount_available).toBe(30);
```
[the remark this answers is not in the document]
r4mmer: Since we iterate on utxos following a different order we are selecting different utxos.

tuliomir (approved):

__tests__/integration/utxo-consolidation.test.js
```diff
@@ -375,21 +378,21 @@ describe('utxo-consolidation routes', () => {
...
     await TestUtils.pauseForWsUpdate();
 
     // Evaluating destination address balance and utxos
     const destinationUtxos = await wallet1.getUtxos({
       filter_address: destinationAddress
     });
-    expect(destinationUtxos.total_amount_available).toBe(70);
+    expect(destinationUtxos.total_amount_available).toBe(90);
     expect(destinationUtxos.total_utxos_available).toBe(1);
-    expect(destinationUtxos.utxos[0].tx_id).toBe(consolidateTx.txId);
```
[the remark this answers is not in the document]
r4mmer: During the first development it was removed, i've added this line back

src/controllers/wallet/wallet.controller.js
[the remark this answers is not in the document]
pedroferreira1: We can continue with this check in the headless in order to merge this PR but we need it in the lib as well, it makes more sense in my opinion

pedroferreira1 (approved):

PR #285, opened 2023-05-18
feat: add parameter to create token and create nft endpoints to set authority addresses

pedroferreira1:
### Acceptance Criteria
- Add parameters in the create token endpoint to choose whether the user wants to create mint/melt authorities, define the addresses to where they will be sent, and if it allows external addresses.
- Add parameters in the create nft endpoint to define the addresses to where these authorities will be sent and if it allows external addresses.

### Security Checklist

### DIFF
--- a/src/api-docs.js
+++ b/src/api-docs.js
@@ -2224,6 +2224,30 @@ const apiDoc = {
                     type: 'string',
                     description: 'Optional address to send the change amount.'
                   },
+                  create_mint: {
+                    type: 'boolean',
+                    description: 'If should create mint authority for the created token. Default is true.'
+                  },
+                  mint_authority_address: {
+                    type: 'string',
+                    description: 'Optional address to send the mint authority output created.'
+                  },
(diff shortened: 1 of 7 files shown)


msbrogli (approved):

src/api-docs.js
```diff
@@ -2224,6 +2224,30 @@ const apiDoc = {
...
+                  },
+                  create_melt: {
+                    type: 'boolean',
+                    description: 'If should create melt authority for the created token. Default is true.'
+                  },
+                  melt_authority_address: {
+                    type: 'string',
+                    description: 'Optional address to send the melt authority output created.'
+                  },
+                  allow_external_melt_authority_address: {
```
r4mmer, line 2247: Why not have unify both flags as `allow_external_authority_address`?

r4mmer (approved):

src/api-docs.js
pedroferreira1: We will discuss this for the official 1.0.0 release

PR #286, opened 2023-05-18
feat: Swap service `update` partialTx

tuliomir:
### Acceptance Criteria
- Adds the option to `update` a proposal with the Atomic Swap Service

### Security Checklist

### DIFF
--- a/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
+++ b/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
@@ -42,7 +42,7 @@ async function buildTxProposal(req, res) {
   if (req.body.lock !== undefined) {
     markAsSelected = req.body.lock;
   }
-  /** @type {{password: string, is_new?: boolean, proposal_id?: string, version?: number}} */
+  /** @type {{password?: string, is_new?: boolean, proposal_id?: string, version?: number}} */
   const serviceParams = req.body.service || {};
 
   const utxos = [];
@@ -137,10 +137,26 @@ async function buildTxProposal(req, res) {
         createdProposalId = proposalId;
       } else if (serviceParams.proposal_id) {
         // Handling the update of an existing proposal with the Atomic Swap Service
+
+        // First, validate if the proposal is already registered with this wallet
+        const listenedProposals = await atomicSwapService.getListenedProposals(req.walletId);
+        if (!listenedProposals.has(serviceParams.proposal_id)) {
+          res.status(404);
+          res.send({
+            success: false,
+            error: 'Proposal is not registered. Register it first through [POST] /register/:proposalId',
+          });
+          return;
+        }
+
+        // Retrieving registered proposal password
+        const requestedProposal = listenedProposals.get(serviceParams.proposal_id);
+        const { password } = requestedProposal;
+
         await atomicSwapService.serviceUpdate(
           {
             proposalId: serviceParams.proposal_id,
-            password: serviceParams.password,
+            password,
             partialTx: proposal.partialTx.serialize(),
             version: serviceParams.version,
           }
--- a/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
+++ b/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
@@ -144,7 +144,7 @@ async function buildTxProposal(req, res) {
           res.status(404);
           res.send({
             success: false,
-            error: 'Proposal is not registered. Register it first through [POST] /register/:proposalId',
+            error: 'Proposal is not registered. Register it first.',
           });
           return;
         }
--- a/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
+++ b/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
@@ -138,7 +138,6 @@ async function buildTxProposal(req, res) {
       } else if (serviceParams.proposal_id) {
         // Handling the update of an existing proposal with the Atomic Swap Service
         await atomicSwapService.serviceUpdate(
-          req.walletId,
           {
             proposalId: serviceParams.proposal_id,
             password: serviceParams.password,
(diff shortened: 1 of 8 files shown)


src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -126,13 +126,42 @@ async function buildTxProposal(req, res) {
...
+      } else if (serviceParams.proposal_id) {
+        // Handling the update of an existing proposal with the Atomic Swap Service
+
+        // First, validate if the proposal is already registered with this wallet
+        const listenedProposals = await atomicSwapService.getListenedProposals(req.walletId);
+        if (!listenedProposals.has(serviceParams.proposal_id)) {
+          res.status(404);
+          res.send({
+            success: false,
+            error: 'Proposal is not registered. Register it first through [POST] /register/:proposalId',
```
r4mmer, line 147: The URL is wrong.
Also, I don't think we should use error messages to point to another API.

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -126,13 +126,42 @@ async function buildTxProposal(req, res) {
...
+          });
+          return;
+        }
+
+        // Retrieving registered proposal password
+        const requestedProposal = listenedProposals.get(serviceParams.proposal_id);
+        const { password } = requestedProposal;
+
+        await atomicSwapService.serviceUpdate(
```
r4mmer, line 156: We do not check for the success response, should we?

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
tuliomir: Removed the URL on [c498c00](https://github.com/HathorNetwork/hathor-wallet-headless/pull/286/commits/c498c00478a141f46e8593dc54df0da66294fc92)

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
tuliomir: The service already checks for the success and throws if it isn't `true`, so we don't need to double check here.

src/services/atomic-swap.service.js
```diff
@@ -74,6 +74,26 @@ const serviceCreate = async (walletId, partialTx, password) => {
...
+ * @returns {Promise<{ success: boolean }>} Returns if the operation was successful
+ */
+const serviceUpdate = async updateParams => {
+  const { success } = await swapService.update(updateParams);
+
+  if (!success) {
+    throw new Error('Unable to update the proposal on the Atomic Swap Service');
+  }
+
+  return { success };
```
r4mmer, line 94: Should we return the success? It will always be true due to the check above, right?

r4mmer (changes requested):

src/services/atomic-swap.service.js
```diff
@@ -74,6 +74,26 @@ const serviceCreate = async (walletId, partialTx, password) => {
...
+ * Creates the proposal on the Atomic Swap Service, handling errors that may occur on the process
+ * @param updateParams Parameters related to the proposal itself
+ * @param {string} updateParams.partialTx Serialized PartialTx
+ * @param {string} updateParams.proposalId Proposal identifier
+ * @param {string} updateParams.password Proposal password on the Atomic Swap Service
+ * @param {number} updateParams.version Proposal version on the Atomic Swap Service
+ * @param {string} [updateParams.signatures] Proposal signatures, if any
+ * @returns {Promise<{ success: boolean }>} Returns if the operation was successful
+ */
+const serviceUpdate = async updateParams => {
```
pedroferreira1, line 87: This is the same method name that you have in the other PR, right? And the other one has a different implementation, with more validations in the beginning.

src/services/atomic-swap.service.js
tuliomir: Improved to returning `Promise<void>` on [4dbf6cd](https://github.com/HathorNetwork/hathor-wallet-headless/pull/286/commits/4dbf6cd55b6b7dd5943ae77654fb6ef1fc952721)

src/services/atomic-swap.service.js
```diff
@@ -29,6 +29,17 @@ const {
  */
 const walletListenedProposals = new Map();
 
+/**
+ * Represents errors related to interactions with the Atomic Swap Service
+ */
+class SwapServiceError extends Error {
```
pedroferreira1, line 35: The same comment as in the other PR.

tuliomir: Added a dedicated `errors` file on [79a66a8](https://github.com/HathorNetwork/hathor-wallet-headless/pull/286/commits/79a66a8202d802f4edc0ccad3a2a7f395b43e5ec)

src/errors.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable max-classes-per-file */
```
pedroferreira1, line 1: Why do you need this?

src/services/atomic-swap.service.js
```diff
@@ -9,6 +9,7 @@ const {
   PartialTxInputData,
   swapService,
 } = require('@hathor/wallet-lib');
+const { SwapServiceError } = require('../errors');
```
pedroferreira1, line 12: Why are you using `require` and not `import`?

pedroferreira1 (approved):

src/errors.js
tuliomir: Our linter complains that each javascript file should have at most one class.

Since this file is dedicated only to custom error classes, it wouldn't make sense to enforce this rule here.

src/services/atomic-swap.service.js
tuliomir: Good point. In this repo, 9 files are using `import` and 21 are using `require`.
It seems like a good issue to raise.

src/errors.js
r4mmer: Why not remove this on the eslint configuration, since it makes more sense to disable this rule altogether rather than make an exception for this file.

r4mmer (approved):

src/errors.js
tuliomir: Looking deeper into this issue, we found that the `wallet lib` also has this rule.
It seems like a good idea to keep it until we decide differently for these projects

PR #288, opened 2023-05-19
feat: Swap service `update` signatures

tuliomir:
### Summary
This PR adds the option for the `/wallet/atomic-swap/tx-proposal/sign` route to also update the Atomic Swap Service with the signatures, besides generating the already expected `txHex`.

### Acceptance Criteria
- Adds the option to `update` the proposal signatures with the Atomic Swap Service

### Security Checklist

### DIFF
--- a/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
+++ b/src/controllers/wallet/atomic-swap/tx-proposal.controller.js
@@ -42,7 +42,7 @@ async function buildTxProposal(req, res) {
   if (req.body.lock !== undefined) {
     markAsSelected = req.body.lock;
   }
-  /** @type {{password: string, is_new?: boolean, proposal_id?: string}} */
+  /** @type {{password?: string, is_new?: boolean, proposal_id?: string, version?: number}} */
   const serviceParams = req.body.service || {};
 
   const utxos = [];
@@ -52,9 +52,16 @@ async function buildTxProposal(req, res) {
     return;
   }
 
-  const proposal = partialTx
-    ? PartialTxProposal.fromPartialTx(partialTx, wallet.storage)
-    : new PartialTxProposal(wallet.storage);
+  // Deserializing the proposal from the partial_tx, if informed, or creating an empty new one
+  let proposal;
+  try {
+    proposal = partialTx
+      ? PartialTxProposal.fromPartialTx(partialTx, wallet.storage)
+      : new PartialTxProposal(wallet.storage);
+  } catch (e) {
+    res.json({ success: false, error: e.message });
+    return;
+  }
 
   if (sendTokens.utxos && sendTokens.utxos.length > 0) {
     try {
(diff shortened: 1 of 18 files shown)


src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -225,10 +225,39 @@ async function signTx(req, res) {
...
 
+    // Updating the service, if a proposal id was informed
+    if (constants.SWAP_SERVICE_FEATURE_TOGGLE && proposalId) {
+      // First, validate if the proposal is already registered with this wallet
+      const listenedProposals = await atomicSwapService.getListenedProposals(req.walletId);
+      if (!listenedProposals.has(proposalId)) {
+        res.status(404);
+        res.send({
+          success: false,
+          error: 'Proposal is not registered. Register it first through [POST] /register/:proposalId',
```
r4mmer, line 241: The URL is wrong, we have some prefix (`/wallet/atomic-swap/...`).
Also, should we have this URL on the error message?

r4mmer (changes requested):

src/services/atomic-swap.service.js
```diff
@@ -74,6 +74,33 @@ const serviceCreate = async (walletId, partialTx, password) => {
...
+    throw new Error('Missing password');
+  }
+
+  const { success } = await swapService.update(updateParams);
+
+  if (!success) {
+    throw new Error('Unable to update the proposal on the Atomic Swap Service');
+  }
+
+  return { success };
```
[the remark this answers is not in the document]
tuliomir: Removed on [f655655](https://github.com/HathorNetwork/hathor-wallet-headless/pull/288/commits/f65565559bb66955c3a3bedde9c7ce6b443454b3).

src/services/atomic-swap.service.js
```diff
@@ -74,6 +74,33 @@ const serviceCreate = async (walletId, partialTx, password) => {
...
+    throw new Error('Missing PartialTx');
+  }
+  if (!updateParams.password) {
+    throw new Error('Missing password');
+  }
+
+  const { success } = await swapService.update(updateParams);
+
+  if (!success) {
+    throw new Error('Unable to update the proposal on the Atomic Swap Service');
```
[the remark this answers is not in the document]
tuliomir: Changed to a more specific `SwapServiceError` on [f655655](https://github.com/HathorNetwork/hathor-wallet-headless/pull/288/commits/f65565559bb66955c3a3bedde9c7ce6b443454b3)

src/services/atomic-swap.service.js
```diff
@@ -29,6 +29,17 @@ const {
  */
 const walletListenedProposals = new Map();
 
+/**
+ * Represents errors related to interactions with the Atomic Swap Service
+ */
+class SwapServiceError extends Error {
```
pedroferreira1, line 35: I think would be better to have a separate file with the errors (and with a TODO to use the errors from the lib when you add it there)

And why don't you create a simple Error class like this?

```
class SwapServiceError extends Error {}
```

r4mmer (approved):

src/services/atomic-swap.service.js
tuliomir: This code was successfully rebased from 286

pedroferreira1 (approved):

PR #289, opened 2023-05-19
fix: Handling invalid `partial_tx` input errors

tuliomir:
### Acceptance Criteria
- Passing an invalid `partial_tx` to `/wallet/atomic-swap/tx-proposal` should result in a treated HTTP response with the error description as a message

### Security Checklist

### DIFF
--- a/__tests__/atomic-swap/tx-proposal-create.test.js
+++ b/__tests__/atomic-swap/tx-proposal-create.test.js
@@ -123,10 +123,10 @@ describe('create tx-proposal api', () => {
         },
       })
       .set({ 'x-wallet-id': walletId });
-    expect(response.status).toBe(400);
+    expect(response.status).toBe(200);
     expect(response.body).toMatchObject({
       success: false,
-      error: 'Invalid serialized partial_tx',
+      error: 'Invalid PartialTx',
     });
   });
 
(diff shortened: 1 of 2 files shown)


src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -52,9 +52,15 @@ async function buildTxProposal(req, res) {
...
-  const proposal = partialTx
-    ? PartialTxProposal.fromPartialTx(partialTx, wallet.storage)
-    : new PartialTxProposal(wallet.storage);
+  let proposal;
+  try {
+    proposal = partialTx
+      ? PartialTxProposal.fromPartialTx(partialTx, wallet.storage)
+      : new PartialTxProposal(wallet.storage);
+  } catch (e) {
+    res.status(400).json({ success: false, error: 'Invalid serialized partial_tx' });
```
luislhl, line 61: Isn't it better to return this only if we are sure the error was caused by an invalid partial_tx?

Maybe checking the exception type to be `SyntaxError` or `UnsupportedScriptError` would achieve this.

__tests__/atomic-swap/tx-proposal-create.test.js
```diff
@@ -113,6 +113,23 @@ describe('create tx-proposal api', () => {
     });
   });
 
+  it('should return 400 with an invalid body', async () => {
```
luislhl, line 116: I think you should just rename this test to replace `400` with `200`.

luislhl (approved):

PR #291, opened 2023-06-05
feat: add authority address parameter to mint/melt endpoints

pedroferreira1:
### Acceptance Criteria
- Add new parameters to set mint/melt authority addresses for mint/melt endpoints

### Security Checklist

### DIFF
--- a/__tests__/integration/melt-tokens.test.js
+++ b/__tests__/integration/melt-tokens.test.js
@@ -1,3 +1,4 @@
+import { transaction as transactionUtils, constants, network, scriptsUtils } from '@hathor/wallet-lib';
 import { TestUtils } from './utils/test-utils-integration';
 import { WALLET_CONSTANTS } from './configuration/test-constants';
 import { WalletHelper } from './utils/wallet-helper';
@@ -308,7 +309,7 @@ describe('melt tokens', () => {
       .send({
         token: tokenA.uid,
         address: await wallet1.getAddressAt(1),
-        amount: 150
+        amount: 100
       })
       .set({ 'x-wallet-id': wallet1.walletId });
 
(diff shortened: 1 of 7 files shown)


__tests__/integration/configuration/test.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
   consoleLevel: process.env.TEST_CONSOLE_LEVEL || 'silly',
 
   // Defines how long tests should wait before consulting balances after transactions
-  wsUpdateDelay: process.env.TEST_WS_UPDATE_DELAY || 1000,
+  wsUpdateDelay: process.env.TEST_WS_UPDATE_DELAY || 2000,
```
pedroferreira1, line 14: I was getting error in the multisig integration test when running in the CI. The error was out of funds to send a transaction, which indicated that the WS update didn't arrive before running the next test. 

After this test, everything succeeded but this shouldn't be considered a final solution, the integration tests deserve a bigger refactor.

__tests__/integration/send-tx.test.js
```diff
@@ -117,7 +117,7 @@ describe('send tx (HTR)', () => {
     expect(response.status).toBe(200);
     expect(response.body.hash).toBeUndefined();
     expect(response.body.success).toBe(false);
-    expect(response.body.error).toContain('Invalid');
+    expect(response.body.error).toContain('Change address is not from the wallet');
```
pedroferreira1, line 120: This should've failed in the previous PR.

__tests__/integration/configuration/test.config.js
tuliomir: Two considerations on this:
1) A change with less impact on the overall test duration would be to add manual delays on the tests that are failing the most ( the ones on MultiSig ). But given this error is happening more often, I agree changing the config is also a valid solution.

2) The best solution for this issue I believe would be using the `external notifications` feature, that was implemented after the integration tests were built. With it, an event would be triggered when the WS update was completed, avoiding the need for arbitrary delay periods.

tuliomir (approved):

__tests__/integration/configuration/test.config.js
pedroferreira1: I will create an issue regarding this problem and describe possible solutions.

r4mmer (approved):

PR #293, opened 2023-06-09
feat: add parameter to create token and create nft endpoints to set authority addresses

pedroferreira1:
### Acceptance Criteria
- Add parameters in the create token endpoint to choose whether the user wants to create mint/melt authorities, define the addresses to where they will be sent, and if it allows external addresses.
- Add parameters in the create nft endpoint to define the addresses to where these authorities will be sent and if it allows external addresses.

### Security Checklist

### DIFF
--- a/src/controllers/wallet/wallet.controller.js
+++ b/src/controllers/wallet/wallet.controller.js
@@ -435,11 +435,33 @@ async function createToken(req, res) {
   const { name, symbol, amount } = req.body;
   const address = req.body.address || null;
   const changeAddress = req.body.change_address || null;
+  const createMint = req.body.create_mint === undefined ? true : req.body.create_mint;
+  const mintAuthorityAddress = req.body.mint_authority_address || null;
+  const allowExternalMintAuthorityAddress = req.body.allow_external_mint_authority_address || false;
+  const createMelt = req.body.create_melt === undefined ? true : req.body.create_melt;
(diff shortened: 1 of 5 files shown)


r4mmer (approved):

src/controllers/wallet/wallet.controller.js
```diff
@@ -435,11 +435,33 @@ async function createToken(req, res) {
   const { name, symbol, amount } = req.body;
   const address = req.body.address || null;
   const changeAddress = req.body.change_address || null;
+  const createMint = req.body.create_mint === undefined ? true : req.body.create_mint;
```
tuliomir, line 438: I think this would be a good place to use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#examples). It would make the code more readable and concise.
```suggestion
  const createMint = req.body.create_mint ?? true;
```

tuliomir (approved):

src/controllers/wallet/wallet.controller.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/293/commits/6bee2632a895c9caadedbabcf014a2ee4e2a9e4e

Issue #292, opened 2023-06-09
tests: Improve stability for integration tests


pedroferreira1:
Sometimes it's important in the integration tests to wait in the websocket for a transaction that was just sent, so the wallet can update its utxos and balance, in order to continue the test sequence. We are currently using a delay but this is causing some issues because sometimes the delay used is not enough and the test fails intermittently.

There are some solutions for that, one is what was done in the wallet lib, which is to await for a specific transaction to arrive in the websocket before continuing the test sequence. We also have the headless wallet external notification plugins that could be used for that.

PR #306, opened 2023-06-16
tests: fix promise left hanging after unit tests finish

pedroferreira1:
### Context

Issue: https://github.com/HathorNetwork/hathor-wallet-headless/issues/292

The following log was found during unit tests

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  TCPSERVERWRAP

... 

  at Test.Object.<anonymous>.Test.serverAddress (node_modules/supertest/lib/test.js:59:33)
      at new Test (node_modules/supertest/lib/test.js:36:12)
      at Object.obj.(anonymous function) [as get] (node_modules/supertest/index.js:25:14)
```

Checking this [issue](https://github.com/ladjs/supertest/issues/520) one of the possibilities was that the supertest server was not being closed after the test.

I used the solution proposed [here](https://github.com/ladjs/supertest/issues/520#issuecomment-436071071)

### Acceptance Criteria
- Close open promises after unit test end
- I decided to keep `forceExit` parameter for a while to understand if there could be other hanging places.

### Security Checklist

### DIFF
--- a/__tests__/test-utils.js
+++ b/__tests__/test-utils.js
@@ -37,8 +37,7 @@ const MULTISIG_DATA = {
   }
 };
 
-const app = createApp();
-const request = supertest(app);
+let request, server;
 
 const httpMock = new MockAdapter(axios);
 
@@ -226,10 +225,26 @@ class TestUtils {
         }
       });
     });
+
+  }
+
+  static startServer() {
+    return new Promise((resolve, reject) => {
+      const app = createApp();
+      server = app.listen(8088, (err) => {
+        if (err) {
+          reject(err);
+        }
+
+        request = supertest.agent(server);
+        resolve();
+      });
+    });
   }
 
   static stopMocks() {
     httpMock.reset();
+    server.close();
     return new Promise(resolve => {
       wsMock.stop(resolve);
     });
--- a/__tests__/test-utils.js
+++ b/__tests__/test-utils.js
@@ -37,7 +37,8 @@ const MULTISIG_DATA = {
   }
 };
 
-let request, server;
+let request; let
+  server;
 
 const httpMock = new MockAdapter(axios);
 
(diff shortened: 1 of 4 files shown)


r4mmer (approved):

__tests__/test-utils.js
```diff
@@ -228,8 +228,23 @@ class TestUtils {
     });
   }
 
+  static startServer() {
+    return new Promise((resolve, reject) => {
+      const app = createApp();
+      server = app.listen(8088, err => {
+        if (err) {
+          reject(err);
```
tuliomir, line 236:
```suggestion
          return reject(err);
```
It's recommended to finish the function execution here, otherwise the supertest agent will also be instantiated and the promise will be also resolved, which might raise exceptions.

__tests__/test-utils.js
```diff
@@ -228,8 +228,23 @@ class TestUtils {
...
+        }
+
+        request = supertest.agent(server);
```
tuliomir, line 239:
```suggestion
        // Ensures the supertest agent will be bound to the correct express port
        request = supertest.agent(server);
```
To understand why we're using this way of implementing the supertest would require a git blame or a search on the supertest's underlying agent API. I think a simple comment, as suggested above, would make it easier for a reader to understand why we chose this implementation.

tuliomir (approved):

__tests__/test-utils.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/306/commits/652c02ab2bc917f8b2b6550bcdbbef883f1e60aa

__tests__/test-utils.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/306/commits/652c02ab2bc917f8b2b6550bcdbbef883f1e60aa

PR #308, opened 2023-06-20
tests: wait for multisig tx to arrive before continuing the test

pedroferreira1:
Issue: https://github.com/HathorNetwork/hathor-wallet-headless/issues/292

### Acceptance Criteria
- Wait multisig transaction before continuing test to prevent out of funds error

### Security Checklist

### DIFF
--- a/__tests__/integration/utils/test-utils-integration.js
+++ b/__tests__/integration/utils/test-utils-integration.js
@@ -807,25 +807,48 @@ export class TestUtils {
     return networkHeight;
   }
 
+  /**
+   * Call headless API to get a transaction from the wallet storage
+   *
+   * @param {string} walletId ID of the wallet to get the transaction from
+   * @param {string} txId ID of the transaction to get
+   *
+   * @returns {Promise<*>} The return is an object with { success: boolean }
+   *                       and the transaction object in case of success
+   */
   static async getTransaction(walletId, txId) {
-    return TestUtils.request
+    const response = await TestUtils.request
       .get(`/wallet/transaction/?id=${txId}`)
       .set(TestUtils.generateHeader(walletId));
+
+    return response.body;
   }
 
+  /**
+   * Waits until the transaction arrives in the wallet ws, or a timeout (optional)
+   * is reached
+   *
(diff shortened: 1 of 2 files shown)


r4mmer (approved):

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -806,4 +806,58 @@ export class TestUtils {
...
+    return new Promise(async (resolve, reject) => {
+      let timeoutHandler;
+      if (timeout) {
+        // Timeout handler
+        timeoutHandler = setTimeout(() => {
+          reject(new Error(`Timeout of ${timeout}ms without receiving the tx with id ${txId}`));
+        }, timeout);
+      }
+
+      while ((await TestUtils.getTransaction(walletId, txId)).success === false) {
```
tuliomir, line 850: When a `timeout` is passed as parameter, I believe this `while` block will continue running even after a `reject` is sent. ( [similar case](https://stackoverflow.com/questions/28896280/why-does-javascript-es6-promises-continue-execution-after-a-resolve) on a SO question )

Since this is an util used only on the integration tests, the effects of these "leaks" will probably not be very impactful, but I think it's important to at least document this behavior, so that future improvements can target this easily.

pedroferreira1: From the SO answer I read this:

> Unless an error is thrown, a function is executed until a return statement or its end is reached. Other code outside of the function can't interfere with that (unless, again, an error is thrown).

So if I just add the `return reject` it should be enough, right?

pedroferreira1: Actually, this is not enough because the `return` is ending the `setTimeout` block, I've done some tests here and double-checked it.

I've done the timeout handling a bit differently, what do you think? https://github.com/HathorNetwork/hathor-wallet-headless/pull/308/commits/09783386d74282140e8fea46e2c0e11270a1aa1a

tuliomir: I think that's a good solution :+1:

tuliomir (approved):

Issue #304, opened 2023-06-16
Multisig for bridge - Inspect transaction with wallet metadata


alexruzenhack:
Related to: https://github.com/HathorNetwork/hathor-wallet-headless/issues/296

PR #315, opened 2023-06-24
chore: add metadata to decode endpoint

alexruzenhack:
### Acceptance Criteria
- extende the decode endpoint `/wallet/decode` with transaction, input, output and balance metadatas

### Security Checklist

Closes: https://github.com/HathorNetwork/hathor-wallet-headless/issues/304

### DIFF
--- a/src/controllers/wallet/wallet.controller.js
+++ b/src/controllers/wallet/wallet.controller.js
@@ -6,14 +6,15 @@
  */
 
 // import is used because there is an issue with winston logger when using require ref: #262
+import { HATHOR_TOKEN_CONFIG } from '@hathor/wallet-lib/lib/constants';
 import logger from '../../logger'; // eslint-disable-line import/no-import-module-exports
 
-const { txApi, walletApi, WalletType, constants: hathorLibConstants, helpersUtils, errors, tokensUtils, PartialTx } = require('@hathor/wallet-lib');
+const { txApi, walletApi, WalletType, constants: hathorLibConstants, helpersUtils, errors, tokensUtils, transactionUtils, PartialTx } = require('@hathor/wallet-lib');
 const { matchedData } = require('express-validator');
 const { parametersValidation } = require('../../helpers/validations.helper');
 const { lock, lockTypes } = require('../../lock');
 const { cantSendTxErrorMessage, friendlyWalletState } = require('../../helpers/constants');
-const { mapTxReturn, prepareTxFunds } = require('../../helpers/tx.helper');
+const { mapTxReturn, prepareTxFunds, getTx } = require('../../helpers/tx.helper');
 const { initializedWallets } = require('../../services/wallets.service');
 const { removeAllWalletProposals } = require('../../services/atomic-swap.service');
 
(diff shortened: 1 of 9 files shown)


src/controllers/wallet/wallet.controller.js
```diff
@@ -347,11 +355,55 @@ async function decodeTx(req, res) {
...
+
+    // True if all the inputs are signed, false otherwise
+    data.completeSignatures = data.inputs.length > 0
+      ? data.inputs.every(input => input.signed) // true until find a false statement
+      : false; // empty data.inputs
+
+    // Get balance
+    const balance = {};
+    { // Default HTR balance.
+      const balanceObj = await req.wallet.getBalance(hathorLibConstants.HATHOR_TOKEN_CONFIG.uid);
```
r4mmer, line 372: This is not the balance of the transaction, it is the balance of the entire wallet.
I think you should use [getTxBalance](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/utils/transaction.ts#L209) instead

src/controllers/wallet/wallet.controller.js
```diff
@@ -347,11 +355,55 @@ async function decodeTx(req, res) {
...
+        authorities: {
+          melt: { available: 0, locked: 0 },
+          mint: { available: 0, locked: 0 },
+        }
+      });
+    }
+
+    // Other token balance if any
+    for (const token of data.tokens) {
+      const balanceObj = await req.wallet.getBalance(token);
```
r4mmer, line 387: Same here

r4mmer (changes requested):

src/controllers/wallet/wallet.controller.js
alexruzenhack: Maybe I have misunderstood this. I thought I would return wallet metadata as well, not only transaction metadata. That's why I return the wallet balance.

r4mmer (approved):

src/helpers/tx.helper.js
```diff
@@ -201,8 +201,42 @@ async function prepareTxFunds(wallet, outputs, inputs, defaultToken = HATHOR_TOK
   };
 }
 
+/**
+ * Best effort to get a transaction. First look up at the storage to get a trasaction
```
tuliomir, line 205:
```suggestion
 * Best effort to get a transaction. First look up at the storage to get a transaction
```

src/helpers/tx.helper.js
```diff
@@ -201,8 +201,42 @@ async function prepareTxFunds(wallet, outputs, inputs, defaultToken = HATHOR_TOK
...
+ * If the transaction is not found in the fullnode, return null.
+ *
+ * @param {HathorWallet} wallet
+ * @param {string} id Hash of the transaction to get data from
+ * @return {Promise<DecodedTx|FullNodeTx|null>} Data from the transaction to get.
+ *    Can be null if both the wallet and fullnode does not contain the tx.
+ *
+ * @see DecodedTx at {@link https://github.com/HathorNetwork/hathor-wallet-lib/blob/bc94221cece2bd6d7b64d971ef30b7d593f07e42/src/new/wallet.js#L1058}
+ * @see FullNodeTx at {@link https://github.com/HathorNetwork/hathor-wallet-lib/blob/bc94221cece2bd6d7b64d971ef30b7d593f07e42/src/wallet/types.ts#L500}
+ */
```
tuliomir, line 216: :star2: Nice reference style!

tuliomir (approved):

PR #326, opened 2023-08-05
chore: Upgrade to NodeJS 18

tuliomir:
The [Api-Docs project](https://github.com/HathorNetwork/rfcs/blob/f4760eb17575e644a1eb91480d7f985493303f6c/projects/hathor-wallet-headless/api-docs-ci.md) requires NodeJS version 16 or higher to run correctly.

Since NodeJS 16 itself is just a few weeks away from its end-of-life on [2023-09-11](https://nodejs.org/en/blog/announcements/nodejs16-eol/), it would be better to upgrade our minimum version straight to NodeJS 18.

NodeJS 16 End of Life

### Acceptance Criteria
- The Headless application should only accept NodeJS 18 or superior
- All test suites must be green

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #327, opened 2023-08-05
feat: Api-Docs server

tuliomir:
- Depends on #326
- Partially implements [the Api-Docs design](https://github.com/HathorNetwork/rfcs/blob/f4760eb17575e644a1eb91480d7f985493303f6c/projects/hathor-wallet-headless/api-docs-ci.md)

### Acceptance Criteria
- The developer should be able to lint the `api-docs` locally
- The developer should be able to visualize the rendered `api-docs` locally

### Usage
To start a local webserver with the rendered docs, use
```sh
npm run docs
```

### Security Checklist

> **Note on new dependencies:** The only direct `dev` dependency added was `@redocly/cli`, as it is our [current Api-Docs website provider](https://wallet-headless.docs.hathor.network/).
>
> However, this dependency alone added another 126 indirect `dev` dependencies, raising our global count from 1050 to 1177.
>
> If we deem more secure, the responsibility of having the `@redocly/cli` dependency globally available could be put on the developer side, with the scripts not working until this dependency is made available to `npm` somehow.


scripts/convert-docs.js
```diff
@@ -0,0 +1,21 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const { writeFile } = require('fs/promises');
+const apiDocs = require('../src/api-docs');
+
+(async () => {
+  // Remove redundant `default` property
+  const docsObj = apiDocs.default || apiDocs;
```
r4mmer, line 13: This may be problematic if we create a default key on the root of api-docs.
It shouldn't be a part of the object since openapi does not have a root default key but we would be ignoring this error.

scripts/convert-docs.js
```diff
@@ -0,0 +1,21 @@
...
+
+  // Remove obsolete properties
+  delete docsObj.components.securitySchemes;
```
r4mmer, line 16: We use this to indicate that it may use an apikey, but if it is obsolete, should we use the way to indicate this instead of dynamically fixing the api-docs before linting?

package.json
```diff
@@ -32,7 +32,12 @@
     "lint_fix": "eslint . --fix",
     "generate_words": "babel-node scripts/generate_words.js",
     "generate_wallets": "babel-node scripts/generate_wallets.js",
-    "get_xpub_from_seed": "babel-node scripts/get_xpub_from_seed.js"
+    "get_xpub_from_seed": "babel-node scripts/get_xpub_from_seed.js",
+    "docs_convert": "babel-node scripts/convert-docs.js",
+    "docs_lint": "npx @redocly/cli lint tmp/api-docs.json",
+    "docs_server": "npx @redocly/cli preview-docs tmp/api-docs.json",
+    "docs_clean": "rm tmp/api-docs.json",
+    "docs": "npm run docs_convert && npm run docs_lint && npm run docs_server && npm run docs_clean"
```
r4mmer, line 40: The new script "converts" the api-docs and saves it to `tmp`, why not use known build folder names like `dist` or `build`?

r4mmer (changes requested):

package.json
tuliomir: The idea for this `json` file is to actually be something temporary, and not incorporated into the application in any form - as it is not today. This is reinforced by having the `docs_clean` command: allowing for easy removal of the file at the end of the local validation.

Since the `tmp` folder is usually present at the development environment because of the integration tests, this seemed like a good output folder.

Should we decide to have this as a more persistent development asset, a `build` or `dist` folder could be created.

scripts/convert-docs.js
tuliomir: Our current workflow involves removing this key manually, that's why it is marked as obsolete.
An option would be to leave it there for the developer to validate, but only remove it when executing the production workflow to keep our current process.

@pedroferreira1, do you have any comment about this?

scripts/convert-docs.js
tuliomir: Actually this is a way of retrieving the main object from the `api-docs.js` file, that is exported like this:

https://github.com/HathorNetwork/hathor-wallet-headless/blob/a793c26ed03b895650089b3f4761a150bed454ba/src/api-docs.js#L3822

The `|| apiDocs` was just a way of protecting this script against failures, but indeed it would muffle a case where this obejct export syntax was changed and cause confusion on the developer. This was fixed on [18fdf61](https://github.com/HathorNetwork/hathor-wallet-headless/pull/327/commits/18fdf61b1ed26b601641009d83fbc73c45e773c4).

scripts/convert-docs.js
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const { writeFile } = require('fs/promises');
+const apiDocs = require('../src/api-docs');
+
+(async () => {
```
r4mmer, line 11: Why are we using async and not the sync version of the `writeFile` api?

scripts/convert-docs.js
r4mmer: I think that if we are excluding the apikey from the security schemes when "building" the api-docs, we should not have the apikey on the docs in the first place.
This would make this script essentially unneeded, so we can just serve the api-docs itself.

scripts/convert-docs.js
tuliomir: Just out of habit, I'd say there's no reason to choose one over the other in this specific case.

scripts/convert-docs.js
pedroferreira1: @r4mmer yes, we have this to indicate we support api key. However, back when I created the first docs page, I didn't find a nice way to make it clear that the api key was optional. Since this is a feature that most users don't use (and maybe don't even know that exists), I decided to remove it from the docs page because it was confusing having it in every API.

We can discuss the best approach here but that the way I've been doing to prevent a misunderstanding.

package.json
pedroferreira1: Why don't you save it in the root project folder? It's just one html file, isn't it?

package.json
```diff
@@ -46,6 +51,7 @@
   "homepage": "https://github.com/HathorNetwork/hathor-wallet-headless#readme",
   "devDependencies": {
     "@babel/eslint-parser": "^7.17.0",
+    "@redocly/cli": "^1.0.0",
```
pedroferreira1, line 54: Are you using this anywhere? I don't see it

scripts/convert-docs.js
```diff
@@ -0,0 +1,21 @@
...
+  // Remove `default` export property
+  const docsObj = apiDocs.default;
+
+  // Remove obsolete properties
```
pedroferreira1, line 15: They are not obsolete, the explanation is in the other comment. We should discuss the best approach for that, maybe we should remove from the js docs, not here in the script.

scripts/convert-docs.js
tuliomir: On [9f2ac73](https://github.com/HathorNetwork/hathor-wallet-headless/pull/327/commits/9f2ac7363b8f9c730d99feeac429be93197554c7) a check was implemented to fetch the API Key configuration directly from the config file at runtime.

No arbitrary removal is necessary anymore.

scripts/convert-docs.js
tuliomir: This feature was re-written with a different approach, this thread is now obsolete.

package.json
tuliomir: I'd say this file still belongs to the `tmp` folder instead of `build`, `dist` or the root folder because the file is not used anywhere else outside of this script's scope.

The `[GET] /docs` endpoint, for example, does not use this generated file, and queries its `api-docs.js` file as a source of truth.

In a future moment, as an expansion of this feature, we could event use the `concurrently` dependency to monitor the server lifespan, watch for changes on the `api-docs.js` file and actually delete the generated `json` file right after the server process is closed.

What do you think about this?

r4mmer (approved):
I still think its strange to "build" the docs, but it works better with the preview-docs command.

scripts/convert-docs.js
tuliomir: On [9f2ac73](https://github.com/HathorNetwork/hathor-wallet-headless/pull/327/commits/9f2ac7363b8f9c730d99feeac429be93197554c7) a check was implemented to fetch the API Key configuration directly from the config file at runtime.

No arbitrary removal is necessary anymore.

src/api-docs.js
```diff
@@ -3843,4 +3839,31 @@ const apiDoc = {
   },
 };
 
-export default apiDoc;
+/**
+ * Generates the Api Docs according to the current configurations of the Headless Wallet
+ */
+async function getApiDocs() {
```
pedroferreira1, line 3845: Why is it `async`?

tuliomir: Since the `getApiDocs()` is expected to fetch the latest configurations for the headless instance, I expect it to eventually grow into a promise, needing to fetch information from some other asynchronous source. I added the `async` as a way to keep it already future-proof in that sense.

However, it's not a needed feature at the moment and can safely be refactored to a synchronous method.

tuliomir: `getApiDocs()` reverted from an `async` function to a regular one on [6f75250](https://github.com/HathorNetwork/hathor-wallet-headless/pull/327/commits/6f75250c7ddb85167f520c4f0ea480a53e283bfa)

pedroferreira1 (approved):

PR #339, opened 2023-09-04
fix: patch the express router to handle async errors

r4mmer:
### Acceptance Criteria
- Handle async rejections on async handlers using express middlewares

### Security Checklist

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -7,24 +7,25 @@ export function patchExpressRouter(router) {
       return originalHandler;
     }
 
-    return function(req, res, next) {
+    function wrappedHandler(req, res, next) {
       const ret = originalHandler.call(this, req, res, next);
       if (ret.catch) {
-        ret.catch((err) => next(err));
+        ret.catch(err => next(err));
       }
       return ret;
-    };
+    }
+    return wrappedHandler;
   }
   const methods = ['get', 'post', 'put', 'delete'];
   for (const method of methods) {
     const original = router[method];
     const wrapper = {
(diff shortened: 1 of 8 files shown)


tuliomir (approved):
The code looks good to me, and a smart solution to the problem. However there are two points that I think we could improve:

**1) A better description of the patch's inner workings**
The idea of creating a patch that automatically handles all rejected promises within the handlers is a good one, but the implementation feels a bit arcane to me. Should someone need to maintain this code later I feel it would be hard to understand.

Maybe a more documented docstring within this PR would be enough to improve this, or comments on the code lines themselves.

**2) Test coverage**
No test has failed before this PR, and no test continues to fail after it.
To me, this indicates that we have a gap in coverage for `async` routes failing. Increasing this gap with dedicated tests I feel would be a task for another PR, not this one since it already fixes what it is proposed to fix.

alexruzenhack (approved):

r4mmer:
For future reference.

The issue found is described in express docs ([reference](https://expressjs.com/en/guide/error-handling.html))
>Errors that occur in synchronous code inside route handlers and middleware require no extra work. If synchronous code throws an error, then Express will catch and process it.
>...
>For errors returned from asynchronous functions invoked by route handlers and middleware, you must pass them to the next() function, where Express will catch and process them.

This means that after our refactor to use async handlers the errors were not being sent to the error handler chain due to how async functions return errors to the sync context that called them.
The solution would just be to catch the error from the async handler and send it to the error handler chain, which is precisely what this PR aims to accomplish.

PR #340, opened 2023-09-05
feat: soft reload

r4mmer:
### Acceptance Criteria

- config reload API
- settings module to reload config module
- refactor all possible places to use settings instead of config.
- analyze changes to the config and act accordingly
- middlewares to handle config errors

### Security Checklist


src/controllers/index.controller.js
```diff
@@ -153,6 +154,7 @@ async function start(req, res) {
         seed,
         multisigData,
         passphrase: req.body.passphrase,
+        allowPassphrase: config.allowPassphrase,
```
r4mmer, line 157: This is because instead of importing the settings module on the helper, I think it should receive the flag as a parameter

src/index.js
```diff
@@ -9,87 +9,93 @@
...
-    console.error(`child process error: ${err.message}`);
-  });
-
-  child.on('disconnect', (code, signal) => {
-    console.log(`child process disconnected from IPC channel with (${code} and ${signal})`);
-    // Killing child just in case it has not yet died.
-    child.kill(); // SIGTERM
+import settings from './settings';
+
+async function startHeadless() {
```
r4mmer, line 17: Needed to encapsulate the logic inside an async method since we need to `await` the settings setup below.
The setup is required to be async as well since it makes a dynamic import which can only be async.

src/index.js
```diff
@@ -9,87 +9,93 @@
...
-  apiKey: config.http_api_key,
-  gapLimit: config.gapLimit,
-  connectionTimeout: config.connectionTimeout,
-  enabled_plugins: config.enabled_plugins,
-});
-
-app.listen(config.http_port, config.http_bind_address, () => {
-  console.log(`Listening on ${config.http_bind_address}:${config.http_port}...`);
-});
+startHeadless();
```
r4mmer, line 101: Here we dispatch the async method without an await because the docker version of node does not support root level awaits, we can await this if we bump the version but this works as intended on all versions.

src/services/wallets.service.js
```diff
@@ -5,6 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+const { HathorWallet } = require("@hathor/wallet-lib");
+
+/**
+ * @type {Map<string, HathorWallet>}
+ */
```
r4mmer, line 12: This is purely for type hints, no functional purpose

src/settings.js
```diff
@@ -0,0 +1,226 @@
...
+ */
+
+/**
+ * @typedef {Object} PluginConfig
+ * @property {string} name Plugin name
+ * @property {string} file Plugin filename inside the plugin directory.
+ */
+
+/**
+ * The configuration object
```
r4mmer, line 31: Maybe we should have this type properly defined elsewhere but since we do not use typescript in this project I went with JSDoc, to help keep track of all variations of the config we should check.

scripts/helpers/wallet-precalculation.helper.js
```diff
@@ -7,7 +7,9 @@
 
 import { promises as fs } from 'fs';
 import { addressUtils, walletUtils } from '@hathor/wallet-lib';
-import config from '../../__tests__/integration/configuration/config-fixture';
+import settings from '../../__tests__/integration/configuration/settings-fixture';
```
glevco, line 10: Is this file related to tests?

glevco (approved):

scripts/helpers/wallet-precalculation.helper.js
r4mmer: Yes, the precalculation helps creating the seeds and precalculated addresses that are used on the tests.
The addresses are pre-calculated because the address derivation is somewhat cpu intensive and was hindering the tests execution times, so we just use a file with all addresses pre-calculated.

If we require more in the future or we want to re-generate the addresses, this file will be used.
The helper also expose method to get the pre-calculated wallet so some tests import it directly.

tuliomir: Some comments here:
- The address derivation actually becomes orders of magnitude more cpu intensive _when running inside the test framework_, not sure why, but this was a good solution.
- I think this file belongs to the `scripts` folder because there may be other cases where it's useful to generate multiple wallets and their addresses in batch
- I agree that it feels weird having a script importing modules from the `tests` module. This indeed deserves more thought to come up with a better organization.

src/settings.js
tuliomir, line 1: Would it be possible for this `settings` module not to depend on anything else than the wallet lib?

Since it is so central to the application I feel it will be easy for a circular dependency to be added in a future maintenance, causing unpredictable behaviors/issues or hard-to-follow processes.

Adding to this, the `_analizeConfig` has two responsibilities right now:
- To identify if the changes are valid
- And to interfere with the application if the settings are valid enough

My suggestion would be to split the responsibilities here:
- The `_analizeConfig` could
  - Identify if there are any _NonRecoverable_ errors with the new config passed as parameter
  - Return an object like `{ nonRecoverable: boolean, shouldStop: boolean, initLib: boolean }`
- Whoever called the analysis function would be responsible for taking action upon this
  - Stopping all the wallets
  - Removing proposals
  - Reinitializing the wallet lib

According to this suggestion the responsible would be the `reloadConfig()` caller: the `controllers/index#reloadConfig()` route handler.

Does this make sense?

src/settings.js
```diff
@@ -0,0 +1,249 @@
...
+  ) {
+    await _stopAllWallets();
+    // change network and server in active configuration
+    hathorLibConfig.setNetwork(newConfig.network);
+    hathorLibConfig.setServerUrl(newConfig.server);
+    return;
+  }
+
+  // Checking for changes in tokenUid, gapLimit and connectionTimeout
+  if (
```
tuliomir, line 208: Should the previous suggestion on this file not be accepted: is there a reason why we shouldn't update simultaneously these properties?

For example, if the `serverUrl` and `gapLimit` properties were changed at the same call, only the `serverUrl` would be updated, and another call would be necessary to update the `gapLimit`.

src/middlewares/config-error-handler.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+function ConfigErrorHandler(err, req, res, next) {
+  if (err instanceof NonRecoverableConfigChangeError) {
+    res.status(200);
+    res.send({ success: false, error: NONRECOVERABLE_ERROR });
+    process.exit(0);
+  }
+
+  if (err instanceof UnavailableConfigError) {
+    res.status(503);
+    res.send({ success: false, error: UNAVAILABLE_ERROR });
```
tuliomir, line 25: If we already sent the response here, it would be better to finish execution here and not call the `next` handler, right?

Otherwise we may end up with `Error: Can't set headers after they are sent` errors on the next error handler.

src/app.js
```diff
@@ -62,6 +65,12 @@ const createApp = () => {
   }
 
   app.use(mainRouter);
+  app.use(ConfigErrorHandler);
+  app.use(ReadonlyErrorHandler);
+  app.use((err, req, res, next) => {
+    console.error(err.stack);
```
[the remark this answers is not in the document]
r4mmer: Well, this is the current behavior, the error handling was just moved from the index route to the `createApp` method.
Maybe we should create an issue to implement this behavior to hide the stack on production environments

src/settings.js
```diff
@@ -0,0 +1,249 @@
...
+  // TODO: analize newConfig to check that we have a good config
+
+  // Checking changes in the fields:
+  // http_post, http_bind_address, http_api_key, consoleLevel, httpLogFormat, enabled_plugins
+  if (
+    oldConfig.http_port !== newConfig.http_port
+    || oldConfig.http_bind_address !== newConfig.http_bind_address
+    || oldConfig.http_api_key !== newConfig.http_api_key
+    || oldConfig.consoleLevel !== newConfig.consoleLevel
+    || oldConfig.httpLogFormat !== newConfig.httpLogFormat
```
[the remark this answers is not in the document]
r4mmer: Because these are configured when creating the express app and we have not tested a way to change it while the app is running.
If we find a way to change this configuration while the app is running we can remove it from here and add a case to change this configuration in the future but since it cannot be changed I think that we should stop the app.
This way the user does not think his config change actually worked when it did not.

src/settings.js
r4mmer: > Would it be possible for this settings module not to depend on anything else than the wallet lib?

We can remove the `initHathorLib` but `initializedWallets`, `removeAllWalletProposals` and `errors` are required.

> Does this make sense?

Yes, it would make the module just be a static analysis of the config.
The issue would be just the many different actions that can be taken from the config values (which require access to the old and new configs) and to set the `_config` singleton after the changes are done.

Maybe we should split the analysis and action methods but I think both of them should remain on the settings module  and be called by `reloadConfig`.

```js
// ...
const actions = await _analizeConfig(oldConfig, newConfig);
await _changeSettings(actions);
// ...
```

Something along these lines

src/settings.js
r4mmer: Well, not in this case, since server/network changes would stop all wallets and the next time a wallet is started it would get the new gap-limit/connectionTimeout/tokenUid.
So effectively it would apply the changes to the gapLimit/connectionTimeout/tokenUid

src/settings.js
```diff
@@ -0,0 +1,249 @@
...
+      !== (newConfig.enabled_plugins && newConfig.enabled_plugins.sort().join('#'))
+  ) {
+    throw new errors.NonRecoverableConfigChangeError();
+  }
+
+  // Checking for changes in seed and multisig fields
+  if (oldConfig.seeds) {
+    let shouldStop = false;
+    for (const [seedKey, oldSeed] of Object.entries(oldConfig.seeds)) {
+      // If the new config has changed a seed, we nee to stop the wallets
```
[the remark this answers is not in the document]
r4mmer: Done

src/middlewares/config-error-handler.middleware.js
r4mmer: Done

__tests__/settings.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+  jest.doMock('../src/config', () => ({
+    __esModule: true,
+    default: config,
+  }));
+
+  await settings.setupConfig();
+  await expect(settings.setupConfig()).rejects.toThrow('Cannot setup the configuration twice');
+
+  expect(settings.getConfig).not.toThrow();
```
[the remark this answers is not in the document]
tuliomir: I suggest doing a double check, maybe throwing an error arbitrarily and verifying if the test still passes. I say this because the [Jest documentation](https://jestjs.io/docs/expect#tothrowerror) indicates that wrapping in a function is mandatory.

tuliomir (approved):
Looks good to me. The only adjustments would be a few unused variables already detected by the `github-code-scanning` bot.

tuliomir (approved):

glevco (approved):

PR #342, opened 2023-09-14
feat: add token to decode when the spent tx is not ours

r4mmer:
### Acceptance Criteria
- When the spent transaction is not from our wallet the token field would come as `undefined` and `JSON.stringify` would remove the token property from the decoded transaction.

### Security Checklist

### DIFF
--- a/src/controllers/wallet/wallet.controller.js
+++ b/src/controllers/wallet/wallet.controller.js
@@ -294,7 +294,7 @@ async function decodeTx(req, res) {
     const tokenIndex = (utxo.token_data & TOKEN_INDEX_MASK) - 1;
     if (txObj.tokens.length > tokenIndex) return txObj.tokens[tokenIndex];
     return undefined;
-  };
+  }
 
   const validationResult = parametersValidation(req);
   if (!validationResult.success) {
--- a/src/controllers/wallet/wallet.controller.js
+++ b/src/controllers/wallet/wallet.controller.js
@@ -287,6 +287,15 @@ async function simpleSendTx(req, res) {
 }
 
 async function decodeTx(req, res) {
+  function getToken(utxo, txObj) {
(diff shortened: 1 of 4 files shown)


src/controllers/wallet/wallet.controller.js
```diff
@@ -319,37 +318,42 @@ async function decodeTx(req, res) {
...
+
     const data = {
+      version: tx.version,
+      type: tx.getType(),
       tokens: tx.tokens,
-      inputs: await tx.inputs.reduce(async (acc, input) => {
-        // the accumulator must be awaited to enforce the sequence processing
-        const results = await acc;
-
+      inputs: await Promise.all(tx.inputs.map(async input => {
```
alexruzenhack, line 334: This `Promise.all` can cause some trouble in the future if the number of inputs is big enough. It would add many async tasks that uses blocking resources to the event loop, and Node may not handle it well. This risk is tied to the `getTx` that may query the fullnode all at once. Can we use a `for await (const of tx.inputs)`? `tx.inputs` should be a an iterable.

r4mmer: Thats a good point, changed to a for loop that awaits each response before going to the next.

alexruzenhack (approved):

luislhl (approved):

luislhl (approved):

src/controllers/wallet/wallet.controller.js
```diff
@@ -288,6 +287,15 @@ async function simpleSendTx(req, res) {
 }
 
 async function decodeTx(req, res) {
+  function getToken(utxo, txObj) {
+    if (utxo.token) return utxo.token;
+    if (utxo.token_data === 0) return HATHOR_TOKEN_CONFIG.uid;
```
luislhl, line 292: Should we have tests for this logic?

PR #347, opened 2023-10-03
chore: upgrade node to LTS and update package-lock

r4mmer:
### Acceptance Criteria
- Upgrade docker image to use the LST version of node, v18
- Upgrade CI and tests to use v18
- drop support for v14 and v16 (reason: both currently reached EOL)
- Use `npm clean-install` on docker build to use package-lock correctly

### Security Checklist


r4mmer:
Too many changes need to be made for tests to work on node v18

PR #349, opened 2023-10-13
feat: create APIs for integrating with nano contracts (method, state, history)

pedroferreira1:
### Acceptance Criteria
- Add support for all APIs used to interact with a nano contract
  - Create a nano contract.
  - Execute a method of a nano contract that already exists.
  - Get the state of a nano contract.
  - Get the history of a nano contract.
  - Get oracle data and signed result.
- Add integration and unit tests for the new APIs
- Adapt integration tests to run with experimental nano contract docker image (it requires the network to be `nano-testnet-alpha`)

### Security Checklist


src/routes/wallet/nano-contracts.routes.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * POST request to create a nano contract
+ * For the docs, see api-docs.js
+ */
+nanoContractRouter.post(
+  '/create',
+  body('blueprint_id').isString(),
+  body('address').isString(),
+  body('data').isObject(),
```
r4mmer, line 81: Is there a logic to the data object?
If there is one, we could create a schema to validate it.

r4mmer (commented):
I know we can register nano contracts on the lib, should we add APIs to manage this?

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -916,6 +916,50 @@ export class TestUtils {
...
+    if (timeoutHandler) {
+      clearTimeout(timeoutHandler);
+    }
+
+    if (timeoutReached) {
+      // We must do the timeout handling like this because if I reject the promise directly
+      // inside the setTimeout block, this while block will continue running forever.
+      throw new Error(`Timeout of ${timeout}ms without confirming the tx with id ${txId}`);
+    }
+  }
```
andreabadesso, line 961:
```suggestion
  static async waitTxConfirmed(walletId, txId, timeout) {
    const timeoutPromise = new Promise((_, reject) =>
      setTimeout(() =>
        reject(
          new Error(`Timeout of ${timeout}ms without confirming the tx with id ${txId}`)
        ),
        timeout
      )
    );

    const waitForConfirmationPromise = async () => {
      let number = await TestUtils.getTransactionConfirmationNumber(walletId, txId);

      while (number < 1) {
        if (timeoutReached) {
          break;
        }
        // Tx not confirmed, wait 1s before trying again
        await delay(1000);

        number = await TestUtils.getTransactionConfirmationNumber(walletId, txId);
      }

      // If the loop finished, the promise is resolved successfully, nothing to return.
    };

    return Promise.race([waitForConfirmationPromise, timeoutPromise]);
  }
```

(untested and just a suggestion for aesthetically pleasing code, Promise.race will either resolve successfully or throw, depending on what resolves or rejects first)

src/controllers/wallet/nano-contracts.controller.js
```diff
@@ -0,0 +1,190 @@
...
+async function executeNanoContractMethodHelper(req, res, isInitialize) {
+  const validationResult = parametersValidation(req);
+  if (!validationResult.success) {
+    res.status(400).json(validationResult);
+    return;
+  }
+
+  const canStart = lock.lock(lockTypes.SEND_TX);
+  if (!canStart) {
+    res.send({ success: false, error: cantSendTxErrorMessage });
```
andreabadesso, line 88:
```suggestion
    // TODO: return status code 423
    res.send({ success: false, error: cantSendTxErrorMessage });
```

andreabadesso (dismissed):

pedroferreira1:
@r4mmer 
> I know we can register nano contracts on the lib, should we add APIs to manage this?

The registration of nano contracts is similar to what we have in the custom tokens. We do it for the clients with UI, so users can know their tokens and to protect them from a visual scam.

The headless doesn't have custom token registration because use cases handle them with ID, then we don't need to register nano contracts in the headless for the same reason.

__tests__/integration/utils/test-utils-integration.js
pedroferreira1: I think the code looks more confusing with the recommendation. As discussed privately, I will keep as it is now.

src/controllers/wallet/nano-contracts.controller.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/349/commits/793bb4125ecc17928ac50874d262d61fdf9bd290

src/routes/wallet/nano-contracts.routes.js
pedroferreira1: Yes, much better: https://github.com/HathorNetwork/hathor-wallet-headless/pull/349/commits/1ef86b607f20f8a283a8449a5a5de2d0649be175

src/api-docs.js
```diff
@@ -3368,6 +3368,734 @@ const defaultApiDocs = {
...
+                              type: 'number',
+                            },
+                            {
+                              type: 'boolean',
+                            },
+                          ],
+                        },
+                        description: 'List of arguments for the initialize method.'
+                      },
+                    }
```
r4mmer, line 3816: We could have a nanoContractData at the beggining of the file and use this on both create and execute.
This will make updating this easier in the future.

```js
const nanoContractData = {
  type: 'object',
  description: 'Data of the initialize method for the nano contract created.',
  properties: {
    actions: {
      type: 'array',
      items: {
        type: 'object',
        required: ['type', 'token', 'amount'],
        properties: {
          type: {
            type: 'string',
            description: 'Type of action: \'deposit\' or \'withdrawal\'.'
          },
          token: {
            type: 'string',
            description: 'Token of the action.'
          },
          amount: {
            type: 'integer',
            description: 'Amount to deposit or withdrawal.'
          },
          address: {
            type: 'string',
            description: 'Required for withdrawal, and it\'s the address to send the token to. For deposit is optional and it\'s the address to get the utxo from.'
          },
          changeAddress: {
            type: 'string',
            description: 'Address to send the change amount. Only used for deposit and it\'s optional.'
          },
        }
      },
      description: 'List of actions for the initialize method.'
    },
    args: {
      type: 'array',
      items: {
        oneOf: [
          {
            type: 'string',
          },
          {
            type: 'integer',
          },
          {
            type: 'number',
          },
          {
            type: 'boolean',
          },
        ],
      },
      description: 'List of arguments for the initialize method.'
    },
  }
};
```

```suggestion
                  data: nanoContractData,
```

andreabadesso (dismissed):

src/api-docs.js
pedroferreira1: Much better, thanks. Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/349/commits/24c0609ced2f5eb72c7beb9fb652acfa7b0303ea

andreabadesso (approved):

PR #350, opened 2023-10-17
chore: Pins Github action versions

tuliomir:
### Acceptance Criteria
- All GitHub Actions dependencies should have their versions pinned

### Security Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -24,10 +24,12 @@ jobs:
         node-version: [14.x, 16.x]
 
     steps:
-      - uses: actions/checkout@v2
+      # https://github.com/actions/checkout/releases/tag/v2.7.0
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        # https://github.com/actions/setup-node/releases/tag/v1.4.6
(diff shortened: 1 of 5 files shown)


.github/workflows/integration-test.yml
```diff
@@ -39,22 +41,25 @@ jobs:
...
+        # https://github.com/actions/upload-artifact/releases/tag/v2.3.1
+        uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2
         with:
           name: test-transacion-logs
           path: tmp
 
       - name: Collect docker logs on failure
         if: failure()
-        uses: jwalton/gh-docker-logs@v1
+        # https://github.com/jwalton/gh-docker-logs/releases/tag/v1.0.0
```
luislhl, line 52: Maybe we should take the opportunity to upgrade this version to the latest (https://github.com/jwalton/gh-docker-logs/releases/tag/v2.2.1). v1.0.0 is a very old version.

Apparently their only breaking change was to use `/bin/sh` instead of `/bin/bash`, but this is configurable, so if we have problems with this we would just have to explicitly set to `/bin/bash` again. (see https://github.com/jwalton/gh-docker-logs/releases/tag/v2.0.0)

luislhl (approved):

r4mmer (approved):

.github/workflows/integration-test.yml
tuliomir: That's a good idea! I believe these updates will be better left as a separate PR, and that could also be implemented in our other repositories as well.

PR #354, opened 2023-11-09
tests: wait first block in integration tests to release genesis reward

pedroferreira1:
### Acceptance Criteria
- We must mine at least one block before starting the integration tests, so the genesis reward is unlocked.
- Add timeout in the method to wait for next block.
- The unit tests were running in parallel and they started to fail because of that. A deeper investigation on this matter must be done but in this PR I decided to just start running them serial, so everything is tested and works fine. A issue was created to handle this: https://github.com/HathorNetwork/hathor-wallet-headless/issues/355

### Security Checklist

### DIFF
--- a/__tests__/integration/utils/test-utils-integration.js
+++ b/__tests__/integration/utils/test-utils-integration.js
@@ -794,6 +794,19 @@ export class TestUtils {
   static async waitNewBlock(currentHeight = null) {
     let baseHeight = currentHeight;
 
+    // This timeout is a protection, so the integration tests
+    // don't keep running in case of a problem
+    // After using the timeout as 120s, we had some timeouts
+    // because the CI runs in a free github runner
+    // so we decided to increase this timeout to 600s, so
+    // we don't have this error anymore
+    const timeout = 600000;
+    let timeoutReached = false;
+    // Timeout handler
+    const timeoutHandler = setTimeout(() => {
+      timeoutReached = true;
+    }, timeout);
+
     if (!baseHeight) {
       baseHeight = await TestUtils.getFullNodeNetworkHeight();
     }
@@ -801,9 +814,17 @@ export class TestUtils {
     let networkHeight = baseHeight;
 
     while (networkHeight === baseHeight) {
+      if (timeoutReached) {
+        break;
+      }
+
       networkHeight = await TestUtils.getFullNodeNetworkHeight();
 
       await delay(1000);
+    }
+
+    if (timeoutReached) {
+      throw new Error('Timeout reached when waiting for the next block.');
     }
 
     return networkHeight;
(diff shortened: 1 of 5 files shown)


__tests__/integration/utils/test-utils-integration.js
```diff
@@ -794,18 +794,42 @@ export class TestUtils {
   static async waitNewBlock(currentHeight = null) {
     let baseHeight = currentHeight;
 
+    // This timeout is a protection, so the integration tests
+    // don't keep running in case of a problem
+    // After using the timeout as 120s, we had some timeouts
+    // because the CI runs in a free github runner
+    // so we decided to increase this timeout to 600s, so
+    // we don't have this error anymore
+    const timeout = 600000;
```
andreabadesso, line 803: Should this be a env variable? Maybe we can get it from the project settings in github

The idea is that if the github runner is slow for some issues on their side, we can just tune it accordingly without the need of a code change

What do you think?

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -794,18 +794,42 @@ export class TestUtils {
...
+    let timeoutReached = false;
+    // Timeout handler
```
andreabadesso, line 805: Improve or remove this comment

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -794,18 +794,42 @@ export class TestUtils {
...
+      if (timeoutReached) {
+        break;
+      }
+
       networkHeight = await TestUtils.getFullNodeNetworkHeight();
 
       await delay(1000);
     }
 
+    // Clear timeout handler
```
andreabadesso, line 826: It is clear enough what the next line is doing, I don't think we need this comment

andreabadesso (changes requested):

tuliomir (approved):

__tests__/integration/utils/test-utils-integration.js
tuliomir: While I like this idea, a code change would always be necessary to modify the `env` variables for an action.

The main difference would be the modified file for adjustments: if we fetch this data from the env variables, we would need to change the `.github/workflows/integration-test.yml` instead of the `__tests__/integration/utils/test-utils-integration.js`

pedroferreira1: Tulio is correct, we would just need to change a different file, the env var wouldn't help for your idea.

As discussed on Slack, if that was the case, we could close this comment.

__tests__/integration/utils/test-utils-integration.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/354/commits/e7c3c00d7a86bf63295451174dc058e38a30b2c2

__tests__/integration/utils/test-utils-integration.js
andreabadesso: As we discussed in slack, we can just fetch it from the github variables and use it in the javascript file

andreabadesso (approved):

PR #356, opened 2023-11-18
Feat: Readonly wallet integration with HSM

tuliomir:
### Acceptance Criteria
- The headless wallet should have an integration with the HSM for a read only wallet

### Security Checklist


src/controllers/hsm/hsm.controller.js
```diff
@@ -0,0 +1,149 @@
...
+
+  // Builds the wallet configuration object
+  const config = settings.getConfig();
+  const walletConfig = {
+    xpub: xPub,
+    password: '123',
+    pinCode: '123',
+    multisig: null,
+    connection: null,
+  };
```
r4mmer, line 102: suggestion: use the `getReadonlyWalletConfig` helper here.

I know it's not necessary, but since this helper is used by the normal start method it will be kept updated.

src/services/wallets.service.js
```diff
@@ -44,8 +54,23 @@ async function stopAllWallets() {
...
+ * Returns true if a wallet id represents a hardware wallet
+ * @param {string} walletId
+ * @returns {boolean} True if this is a hardware wallet
+ */
+function isHardwareWallet(walletId) {
+  if (!initializedWallets.has(walletId)) {
+    return false;
+  }
+
+  return hardWalletIds.has(walletId);
```
r4mmer, line 67:
```suggestion
  return initializedWallets.has(walletId) && hardWalletIds.has(walletId);
```

src/controllers/hsm/hsm.controller.js
```diff
@@ -0,0 +1,149 @@
...
+      message: validationObj.reason,
+    });
+    return;
+  }
+
+  // Obtains this wallet's xPub from the HSM
+  const xPub = await hsmService.getXPubFromKey(connectionObj, hsmKeyName, {
+    isReadOnlyWallet: true,
+  });
+  await hsmService.hsmDisconnect();
```
r4mmer, line 92: question: Do we have a test to check that this method is called?

If this is not called the hsm lock will not be unlocked and any subsequent calls to HSM methods will fail.
From what I can see there are some returns that could be problematic in this route.

src/services/wallets.service.js
tuliomir: Done on [519970f](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/519970fb671308993a0b28efd0cf9633440081c4)

src/controllers/hsm/hsm.controller.js
tuliomir: Always good to reuse existing helper methods. Done on [519970f](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/519970fb671308993a0b28efd0cf9633440081c4).

src/controllers/hsm/hsm.controller.js
tuliomir: Good catch! Added new tests and fixed the leaking connections on [f503f41](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/f503f41a05a401ef2a6ff21306fee4cabc21c541).

r4mmer (approved):

src/services/hsm.service.js
```diff
@@ -0,0 +1,203 @@
...
+const { lock, lockTypes } = require('../lock');
+const { hsmBusyErrorMessage } = require('../helpers/constants');
+const { HsmError } = require('../errors');
+
+/**
+ * Connects to the HSM and returns the connection object
+ * @returns {Promise<Hsm>}
+ */
+async function hsmConnect() {
+  // Ensures that only one connection will be open at any given moment
```
pedroferreira1, line 20: Can't you reuse the same connection to start other wallets?

src/services/hsm.service.js
```diff
@@ -0,0 +1,203 @@
...
+  return hsm.connect(hsmConnectionOptions);
+}
+
+/**
+ * Disconnects the headless wallet from the HSM.
+ * The connection object is a singleton, so it must not be informed as parameter.
+ * @returns {Promise<void>}
+ */
+async function hsmDisconnect() {
+  await hsm.disconnect();
```
pedroferreira1, line 46: Can it be disconnected somehow in the middle of an execution? How do you know that you lost the session connection?

src/services/hsm.service.js
```diff
@@ -0,0 +1,203 @@
...
+   * Derivation will be made on
+   * BIP code / Coin / Account /  Change
+   *    m/44' / 280' /      0' /       0
+   *        1 /    2 /      3  /       4
+   */
+
+  // Derivation 1: Bip Code
+  const derivationVersion = options?.version
+    ? options.version
+    : hsm.enums.VERSION_OPTIONS.BIP32_HTR_MAIN_NET;
```
pedroferreira1, line 126: Why don't you get the network in the config, so you can use the default as testnet here, in case this is the configuration?

src/services/hsm.service.js
```diff
@@ -0,0 +1,203 @@
...
+  returnObject.isValidXpriv = true;
+  return returnObject;
+}
+
+/**
+ * Derivates an HTR xPriv key from a given xPriv key
+ * @param {Object} hsmConnection
+ * @param {string} hsmKeyName
+ * @param {Object} [options]
+ * @param {boolean} [options.isReadOnlyWallet] If true, derivation will stop at the account level
```
pedroferreira1, line 99: Can you start a non read only wallet?

src/controllers/hsm/hsm.controller.js
```diff
@@ -0,0 +1,154 @@
...
+
+  // Builds the wallet configuration object
```
pedroferreira1, line 105: From this line below it's exactly the same as in the `/start` of the other wallets (except for the promise resolve). Can we make this refactor, so we don't duplicate code?

If we change anything from the `/start` endpoint we would need to remember to change it here as well, this is not good.

pedroferreira1 (commented):
Isn't it missing an update in the `config.js.template` and in the `config.js.docker` with the new optional configs for hsm?

package-lock.json
```diff
@@ -1126,6 +1126,11 @@
         "kuler": "^2.0.0"
       }
     },
+    "@dinamonetworks/hsm-dinamo": {
```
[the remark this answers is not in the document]
tuliomir: No, they don't. 🤩 Nice, isn't it?

src/services/hsm.service.js
tuliomir: For future implementations we may want to execute the fourth derivation for calculating the change addresses on the HSM. This flag allows this alternative.

src/services/hsm.service.js
tuliomir: Yes, even though it's unlikely, an error can occur in the middle of an execution. In that case we must have good error handling because the exception must be raised and handled by the controller.

This `disconnect()` method is a no-op in case the connection has already been closed or dropped, so it's safe to call on the controller's exception handling.

src/services/hsm.service.js
tuliomir: While using the dinamo lib we currently can't. [A comment](https://github.com/HathorNetwork/rfcs/pull/67#discussion_r1462591197) was made on the design about this specific point.

src/controllers/hsm/hsm.controller.js
tuliomir: This is a good idea. The refactor will be started on #392 and used here.

src/services/wallets.service.js
```diff
@@ -12,16 +12,26 @@ const { removeAllWalletProposals } = require('./atomic-swap.service');
  */
 const initializedWallets = new Map();
 
+/**
+ * @type {Map<string, String>}
+ */
+const hardWalletIds = new Map();
```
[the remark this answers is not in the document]
tuliomir: Renamed on [2cce965](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/2cce965472b652cdb9dfd16e2d45f683badfc036).

src/services/wallets.service.js
```diff
@@ -12,16 +12,26 @@ const { removeAllWalletProposals } = require('./atomic-swap.service');
  */
 const initializedWallets = new Map();
 
+/**
+ * @type {Map<string, String>}
```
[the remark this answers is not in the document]
tuliomir: A better description was added on [2cce965](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/2cce965472b652cdb9dfd16e2d45f683badfc036)

src/services/hsm.service.js
tuliomir: Actually, one of the starting conditions didn't have a protection for this. It was added on [6756809](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/675680910105fbd54ad684a44c88016d9db1b20c).

src/services/hsm.service.js
tuliomir: Added on [b332161](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/b3321613176a720d172eb4f12898e34f345ad84e)

src/controllers/hsm/hsm.controller.js
tuliomir: Refactor completed, the helper method is already being used instead of the duplicated code.

tuliomir:
> Isn't it missing an update in the `config.js.template` and in the `config.js.docker` with the new optional configs for hsm?

Good point! On [ecbb17b](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/ecbb17b5c453425ea3dcbe754706888d97fb0b78) added the default configs and implemented a better user experience in case these configs were not informed when calling an HSM endpoint. Also added tests for this new error message.

src/services/hsm.service.js
```diff
@@ -0,0 +1,215 @@
...
+    success: true,
+    htrKeyName: childKeyNames.HTR_CKD_CHANGE_KEYNAME,
+  };
+}
+
+/**
+ * Derivates an HTR wallet xPub string from a given HSM xPriv key
+ * @param {Object} hsmConnection
+ * @param {string} hsmKeyName
+ * @param {Object} [options]
```
pedroferreira1, line 196: What are the options here?

src/services/hsm.service.js
pedroferreira1: I didn't understand the reason for that, to be honest, when we want to derive until the change path and when we want to derive until the account path?

src/services/hsm.service.js
tuliomir: As discussed on the comment above, this parameter was removed.

src/services/hsm.service.js
tuliomir: As we've talked, I added some clarifications to the method parameters regarding which levels are derived from the inputs. Done on [2bf7108](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/commits/2bf7108313b675fabdeb32f7a4d60bdd853d86df).

pedroferreira1 (approved):

PR #358, opened 2023-11-24
feat: correctly set await to wait for new tx received before continuing the tests

pedroferreira1:
### Acceptance Criteria
- Include here all things that this PR should solve

### Security Checklist


__tests__/melt-tokens.test.js
```diff
@@ -5,11 +5,11 @@ const walletId = 'stub_melt_tokens';
 
 // TODO: Add a fixture to melt token 01
 describe('melt-tokens api', () => {
-  beforeAll(async () => {
+  beforeEach(async () => {
```
pedroferreira1, line 8: I was having some problems after the first test that executed the melt with success, the others were failing because the wallet didn't have tokens to melt.

__tests__/p2sh/tx-proposal-melt-tokens.test.js
```diff
@@ -77,7 +77,7 @@ describe('melt-tokens tx-proposal api', () => {
       .set({ 'x-wallet-id': walletId });
     expect(response.status).toBe(200);
     expect(response.body.success).toBe(false);
-    expect(response.body.error).toEqual('Not enough tokens to melt: 1000000000000 requested, 200 available');
+    expect(response.body.error).toContain('Not enough tokens to melt: 1000000000000 requested');
```
pedroferreira1, line 80: Using the best utxo selection algorithm we don't return the amount available when we don't have enough funds. Should we create an issue to improve this in the lib?

__tests__/p2sh/tx-proposal-mint-tokens.test.js
```diff
@@ -81,7 +81,7 @@ describe('mint-tokens tx-proposal api', () => {
       })
       .set({ 'x-wallet-id': walletId });
     expect(response.status).toBe(200);
-    expect(response.body.error).toEqual('Not enough HTR tokens for deposit: 10000000000 required, 89600 available');
+    expect(response.body.error).toContain('Not enough HTR tokens for deposit: 10000000000 required');
```
pedroferreira1, line 84: Using the best utxo selection algorithm we don't return the amount available when we don't have enough funds. Should we create an issue to improve this in the lib?

__tests__/integration/multisig.test.js
```diff
@@ -658,22 +654,6 @@ describe('send tx (HTR)', () => {
             txId: expect.any(String),
             value: 2,
           },
-          {
```
pedroferreira1, line 661: This is the tx hex affected by the change in the lib, now that we are using the best utxo selection algorithm.

Previously, this would randomly fail depending on the utxo selected. Now we will always select the same (the best)

__tests__/integration/utils/test-utils-integration.js
```diff
@@ -855,38 +865,109 @@ export class TestUtils {
...
 
-      while ((await TestUtils.getTransaction(walletId, txId)).success === false) {
-        if (timeoutReached) {
-          break;
-        }
-        // Tx not found, wait 1s before trying again
-        await delay(1000);
+    let result = await TestUtils.getTransaction(walletId, txId);
+
+    while (result.success === false || result.processingStatus !== 'finished') {
```
pedroferreira1, line 879: I wanted to use the ENUM from the lib but we are not exporting the full types file, so I couldn't. Maybe we should export all types.

__tests__/melt-tokens.test.js
tuliomir: This is curious. The unit tests were supposed to have no side effects because of the lack of interaction with the blockchain.

Maybe it has to do with the new storage scheme building local indexes on the app?

This looks inoffensive to me, but I believe it's a flag that maybe we're having unwanted side effects from our methods.

__tests__/p2sh/tx-proposal-melt-tokens.test.js
tuliomir: Maybe not on the lib, but on the headless itself.

I think it's not the responsibility of utxo selection method to inform a token balance, but it's a useful information for the end user of the headless.

tuliomir (approved):

PR #361, opened 2023-12-05
chore: allow configuration of index-limit scanning policy

r4mmer:
### Acceptance Criteria

- Allow starting a wallet with the new `index-limit` address scanning policy
- Allow a per-wallet configuration of the `gap-limit` address scanning policy

### Security Checklist


__tests__/wallet-config.test.js
```diff
@@ -0,0 +1,80 @@
...
+    await expect(wallet.storage.getScanningPolicy()).resolves.toBe('index-limit');
+    await expect(wallet.storage.getIndexLimit()).resolves.toMatchObject({
+      startIndex: 5,
+      endIndex: 10,
+    });
+    await expect(wallet.storage.store.addressCount()).resolves.toBe(6);
+
+    response = await TestUtils.request
+      .post('/wallet/config/index-limit/last-index')
+      .send({ index: 14 })
```
alexruzenhack, line 72: [❓ question] In this case it has the same effect of `load-more-addresses`?

src/api-docs.js
```diff
@@ -60,6 +60,25 @@ const defaultApiDocs = {
...
+                  },
+                  policyStartIndex: {
+                    type: 'number',
+                    description: 'Load addresses starting from this index. Only used when scanPolicy is set to \'index-limit\'.',
+                    default: 0,
+                  },
+                  policyEndIndex: {
+                    type: 'number',
+                    description: 'Stop loading addresses at this index. Only used when scanPolicy is set to \'index-limit\'. Defaults to policyStartIndex',
+                  },
```
alexruzenhack, line 81: [💬 comment] This is interesting, I thought that a `index-limit` policy were a complement to the `gap-limit` policy, like an upper bound limit.

src/api-docs.js
```diff
@@ -3973,6 +3992,199 @@ const defaultApiDocs = {
...
+        parameters: [
+          {
+            name: 'x-wallet-id',
+            in: 'header',
+            description: 'Define the key of the corresponding wallet it will be executed the request.',
+            required: true,
+            schema: {
+              type: 'string',
+            },
+          },
```
alexruzenhack, line 4127: [💬  comment] As this is a `js` document, why not parameterize this chunk of code that is repeating over the document?

src/controllers/index.controller.js
```diff
@@ -112,13 +114,57 @@ async function start(req, res) {
...
+  }
+
   let walletConfig;
   if ('xpubkey' in req.body) {
     try {
       // starting a readonly wallet
       walletConfig = getReadonlyWalletConfig({
         xpub: req.body.xpubkey,
         multisigData,
+        scanPolicy: scanPolicyData,
```
alexruzenhack, line 167: [⭕ verify] The `scanPolicyData` should be validated inside the `getReadlongWalletConfig` because it can be evaluated to non integer value after `parseInt`.

src/controllers/index.controller.js
```diff
@@ -155,6 +201,7 @@ async function start(req, res) {
         multisigData,
         passphrase: req.body.passphrase,
         allowPassphrase: config.allowPassphrase,
+        scanPolicy: scanPolicyData,
```
alexruzenhack, line 204: [⭕ verify] The `scanPolicyData` should be validated inside the `getReadlongWalletConfig` because it can be evaluated to non integer value after `parseInt`.

src/helpers/wallet.helper.js
```diff
@@ -44,6 +44,7 @@ export function initHathorLib(config) {
 export function getReadonlyWalletConfig({
   xpub,
   multisigData = null,
+  scanPolicy = null,
 }) {
   // Previous versions of the lib would have password and pin default as '123'
   // We currently need something to be defined, otherwise we get an error when starting the wallet
```
alexruzenhack, line 50: [🚫 no entry] It is with the default value. Should we remove in this PR or another PR will solve it?

src/helpers/wallet.helper.js
```diff
@@ -83,6 +86,7 @@ export function getWalletConfigFromSeed({
     password: '123',
     pinCode: '123',
```
alexruzenhack, line 87: [🚫 no entry] It is with the default value. Should we remove in this PR or another PR will solve it?

src/helpers/wallet.helper.js
```diff
@@ -62,6 +64,7 @@ export function getWalletConfigFromSeed({
   multisigData = null,
   passphrase = null,
   allowPassphrase = false,
+  scanPolicy = null,
```
alexruzenhack, line 67: [❓ question] Shouldn't the `scanPolicy` be validated here?

src/helpers/wallet.helper.js
```diff
@@ -83,6 +86,7 @@ export function getWalletConfigFromSeed({
     password: '123',
     pinCode: '123',
     multisig: multisigData,
+    scanPolicy,
```
alexruzenhack, line 89: [❓ question] Shouldn't the `scanPolicy` be validated here?

__tests__/wallet-config.test.js
r4mmer: They are the same funcionality but implemented differently, instead of specifying that we want to load X more addresses it makes it load addresses until we reach index X (inclusive).
In this example we started with the addresses `[5, 10]` loaded and then we set last index to 14, making it `[5, 14]` (10 addresses total), if we had used `load-more-addresses` with 14 it would end up with `[5, 24]` (20 addresses total)

src/api-docs.js
r4mmer: This is being done in #362

src/controllers/index.controller.js
r4mmer: It it is not an integer, it would return `NaN`, which is a falsy value and the or cases (`||`) would prevent non integers.

Although I still believe we should improve validation on this route, which should be done with `express-validator` a dependency already used on other routes.
The validator would check for int values and return an error message to the user if invalid data was received.
This is outside the scope of this PR, but I think we should create validators for all routes that receive user input.

src/helpers/wallet.helper.js
r4mmer: This warning is only for the pin/password, the `scanPolicy` as `null` would use the default i.e. `gap-limit`

src/helpers/wallet.helper.js
r4mmer: We validate on the route.
Any unsupported policy returns `{"success":false,"message":"Address scanning policy bla is not supported."}`
When using gap-limit we validate the `gap-limit` parameter and when using index-limit we validate the start and end limit parameters.

__tests__/wallet-config.test.js
alexruzenhack: Oh, understood. Thank you!

src/api-docs.js
alexruzenhack: Got it, thank you!

src/api-docs.js
alexruzenhack: Nice!

src/controllers/index.controller.js
alexruzenhack: Oh! You right. I miss the `||` operator.

src/helpers/wallet.helper.js
alexruzenhack: Great! Thank you!

alexruzenhack (approved):

package.json
```diff
@@ -12,7 +12,7 @@
     "@babel/node": "^7.13.0",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@babel/preset-env": "^7.13.5",
-    "@hathor/wallet-lib": "^1.0.3",
+    "@hathor/wallet-lib": "^1.0.4",
```
andreabadesso, line 15: Since we're using "^", npm will already install 1.0.4 (if it's the latest)

Actually, it will install any version that is <2.0.0

Should we fix the version here? 

```suggestion
    "@hathor/wallet-lib": "1.0.4",
```

andreabadesso (approved):

package.json
r4mmer: I think we should lock the wallet-lib version, but i'll make this on another PR.
There is a ktlo to start using some best practices, i'll add this on that PR.

PR #362, opened 2023-12-11
Chore: Lint apidocs

tuliomir:
### Acceptance Criteria
- API Docs should be linted at CI

### Notes

> [!TIP] 
> Since the change on the `src/api-docs.js` file is big, it can be useful to review each commit at a time. Each one of them address a single linting rule, so it's easier to track the context of each change.

### Security Checklist


src/api-docs.js
```diff
@@ -11,17 +11,17 @@ const defaultApiDocs = {
...
+      name: 'MIT',
+      url: 'https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/LICENSE'
+    },
   },
-  produces: ['application/json'],
   components: {},
-  security: [
-    {
-      ApiKeyAuth: [],
-    }
```
tuliomir, line 20: ApiKeyAuth is only added if it is present on the `config.js` file

.redocly.lint-ignore.yaml
```diff
@@ -0,0 +1,54 @@
+# This file instructs Redocly's linter to ignore the rules contained for specific parts of your API.
+# See https://redoc.ly/docs/cli/ for more information.
+tmp/api-docs.json:
+  operation-4xx-response:
+    - '#/paths/~1start/post/responses'
```
r4mmer, line 5: Why ignore all of these?

r4mmer (commented):
Do you think this is a good oportunity to create reusable schemas?
For instance, all wallet routes use the `x-wallet-id` header and have some errors for missing or incorrect wallet-id, maybe we can define them once and just reuse on each route.

pedroferreira1 (approved):

r4mmer (approved):

PR #363, opened 2023-12-11
feat: add support for timelock in the send tx API

pedroferreira1:
### Motivation

A use case needs to create transactions with timelock using the headless.

### Acceptance Criteria
- Add support for timelock in the `send-tx` API.

### Security Checklist


r4mmer (approved):
Should we have a test for this?

tuliomir (approved):
I agree that we should have a test for this, preferably an integration test.

tuliomir (approved):

r4mmer (approved):

PR #364, opened 2023-12-12
feat: Upload the api-docs on CI

tuliomir:
Finishes implementation of the [Api Docs CI](https://github.com/HathorNetwork/rfcs/blob/master/projects/hathor-wallet-headless/api-docs-ci.md) RFC.

When a commit on the `release` branch receives a valid version `tag`, this workflow automatically uploads the generated apidocs to the official apidocs website.

### Acceptance Criteria
- Automates the updating of the official [Headless Wallet apidocs](https://wallet-headless.docs.hathor.network/)

### Alternatives
Instead of creating a whole new workflow, we could have reused the existing `apidocs.yml` from #362 to generate the Api-Docs and lint them. The only steps left would be caching the results and uploading them to S3. The different approaches to this had some drawbacks:

##### Uploading on the same job
We could add the version validation to the `lint` job and upload the S3 right there. However this would make a `lint` job also have an eventual side-effect of changing our official documentation. This seemed like an obfuscated workflow that would raise yellow flags on future maintenance.

##### Creating a new job to upload
This would have a better separation of concerns for the jobs. However, every PR from now on would have a skipped job on its validations, except for the `release` ones. It is also a possibility that the action would not be executed on the `pull_request` trigger and would have to be manually re-run, which would make the automation criteria fail.

### DIFF
--- a/.github/workflows/apidocs-upload.yml
+++ b/.github/workflows/apidocs-upload.yml
@@ -62,7 +62,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies
-        run: npm install
+        run: npm ci
 
       # We use the config module to configure the apidocs generation
       - name: Create config module
--- a/.github/workflows/apidocs-upload.yml
+++ b/.github/workflows/apidocs-upload.yml
@@ -81,5 +81,5 @@ jobs:
       # Copy the document to S3
       - name: Upload apidocs.json to S3
         run: |
-          aws s3 cp ./tmp/apidocs.json ${{ secrets.AWS_APIDOCS_BUCKET_NAME }}
+          aws s3 cp ./tmp/api-docs.json ${{ secrets.AWS_APIDOCS_BUCKET_NAME }}
       
--- a/.github/workflows/apidocs-upload.yml
+++ b/.github/workflows/apidocs-upload.yml
@@ -21,9 +21,13 @@ jobs:
   
           if [[ "${{ github.ref }}" = refs/tags/* ]]; then
             version="${ref#refs/tags/}"
-            # This patterns accepts "v1.0.0" , "v2.4.6". Rejects "v1.0.0-rc1".
+
+            # Pattern that accepts "v1.0.0" , "v2.4.6". 
             if [[ "$version" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
               result=isRelease
+            # Pattern that accepts "v1.0.0-rc1".
+            elif [[ "$version" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+$ ]]; then
+              result=isReleaseCandidate
             else
               result=isNotRelease
             fi
@@ -38,7 +42,9 @@ jobs:
   upload-apidocs:
     runs-on: ubuntu-22.04
     needs: check-version
-    if: ${{ needs.test_version.outputs.release_status_output == 'isRelease' }}
+    env:
+      release_status: ${{ needs.test_version.outputs.release_status_output }}
+    if: ${{ env.release_status == 'isRelease' || env.release_status == 'isReleaseCandidate' }}
     timeout-minutes: 10
 
     strategy:
@@ -83,5 +89,9 @@ jobs:
       # Copy the document to S3
       - name: Upload apidocs.json to S3
         run: |
-          aws s3 cp index.html ${{ secrets.AWS_APIDOCS_BUCKET_NAME }}
+          if [[ "${{ env.release_status }}" == "isReleaseCandidate" ]]; then
+            aws s3 cp index.html ${{ secrets.AWS_APIDOCS_BUCKET_NAME_DEV }}
+          else
+            aws s3 cp index.html ${{ secrets.AWS_APIDOCS_BUCKET_NAME_PROD }}
+          fi
       
--- a/.github/workflows/apidocs-upload.yml
+++ b/.github/workflows/apidocs-upload.yml
@@ -49,8 +49,8 @@ jobs:
       # https://github.com/actions/checkout/releases/tag/v4.0.0
       - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
       - name: Use Node.js ${{ matrix.node-version }}
-        # https://github.com/actions/setup-node/releases/tag/v3.8.1
-        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d
+        # https://github.com/actions/setup-node/releases/tag/v4.0.0
+        uses: actions/setup-node@8f152de45cc393bb48ce5d89d36b731f54556e65
         with:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/apidocs-upload.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    strategy:
+      matrix:
+        node-version: [18.x]
+
+    steps:
+      # https://github.com/actions/checkout/releases/tag/v4.0.0
+      - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
+      - name: Use Node.js ${{ matrix.node-version }}
+        # https://github.com/actions/setup-node/releases/tag/v3.8.1
```
luislhl, line 52: Why not use the newer v4?

.github/workflows/apidocs-upload.yml
```diff
@@ -0,0 +1,85 @@
...
+          fi
+  
+          echo $result
+          echo "version_status_output=$result" >> $GITHUB_OUTPUT
+
+  # Initiate upload only if the version is valid
+  upload-apidocs:
+    runs-on: ubuntu-20.04
+    needs: check-version
+    if: ${{ needs.test_version.outputs.release_status_output == 'isRelease' }}
```
luislhl, line 41: We should consider that we will have a `dev` environment as well. We need to deploy both the `dev` and `prod`, according to what branch/tag trigger the build.

I would say we should deploy to `dev` when there is a commit in the `master` branch, and to `prod` using your current logic.

.github/workflows/apidocs-upload.yml
```diff
@@ -0,0 +1,85 @@
...
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+
+  # First check if this version is a valid public release
+  check-version:
+    runs-on: ubuntu-20.04
```
luislhl, line 12: Should we use a newer version of ubuntu?

.github/workflows/apidocs-upload.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: npm run docs_convert
+      - name: Lint the apidocs json
+        run: npm run docs_lint
+
+      # Configure AWS credentials
+      - name: Set up AWS CLI
+        # https://github.com/aws-actions/configure-aws-credentials/releases/tag/v4.0.1
+        uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
luislhl, line 77: We should preferrably make it assume a Role, instead of using credentials.

I can help you with configuring the Role.

.github/workflows/apidocs-upload.yml
tuliomir: Updated on [eb6a3d8](https://github.com/HathorNetwork/hathor-wallet-headless/pull/364/commits/eb6a3d8d111b647d9563277446a37942e36f615f)

.github/workflows/apidocs-upload.yml
tuliomir: Updated on [f3a3fd5](https://github.com/HathorNetwork/hathor-wallet-headless/pull/364/commits/f3a3fd5fef5e52c5702397a7ebabdc1ea5832284)

.github/workflows/apidocs-upload.yml
tuliomir: I'm under the impression that uploading at every commit to `master` would be confusing to our current deployment structure.

On [6c051a7](https://github.com/HathorNetwork/hathor-wallet-headless/pull/364/commits/6c051a71aaf053e03754eadf005e35c9425f7cc4) I implemented a system where the _release candidates_ are sent to a staging environment. What do you think about this approach?

.github/workflows/apidocs-upload.yml
tuliomir: Done on [b807bdf](https://github.com/HathorNetwork/hathor-wallet-headless/pull/364/commits/b807bdfde8a4d0f108e8788888680014b95f0b6c)

luislhl (dismissed):

src/api-docs.js
```diff
@@ -203,34 +228,20 @@ const defaultApiDocs = {
...
-                  },
-                  'invalid-wallet-id': {
-                    summary: 'Wallet id parameter is invalid',
-                    value: { success: false, message: 'Invalid wallet-id parameter.' }
-                  },
                 },
               },
             },
           },
+          400: { $ref: '#/components/responses/MissingWalletIdError' },
```
pedroferreira1, line 235: Are we returning 400 in case of missing wallet id?

.redocly.lint-ignore.yaml
```diff
@@ -0,0 +1,9 @@
+# This file instructs Redocly's linter to ignore the rules contained for specific parts of your API.
+# See https://redoc.ly/docs/cli/ for more information.
+tmp/api-docs.json:
+  operation-4xx-response:
```
pedroferreira1, line 4: Why are you ignoring these?

.github/workflows/apidocs.yml
```diff
@@ -0,0 +1,39 @@
+name: apidocs
+on:
+  pull_request:
+    branches:
+      - master
+      - release-candidate
+      - release
+jobs:
+  lint:
+    runs-on: ubuntu-20.04
```
pedroferreira1, line 10: Should we use 22.04, as in the other workflow?

.github/workflows/apidocs.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      # https://github.com/actions/checkout/releases/tag/v4.0.0
+      - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
+      - name: Use Node.js ${{ matrix.node-version }}
+        # https://github.com/actions/setup-node/releases/tag/v3.8.1
+        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm install
```
pedroferreira1, line 26: We should use `npm ci` for CI runners intead of `npm install`

.github/workflows/apidocs-upload.yml
```diff
@@ -0,0 +1,98 @@
...
+    steps:
+      # https://github.com/actions/checkout/releases/tag/v4.0.0
+      - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
+      - name: Use Node.js ${{ matrix.node-version }}
+        # https://github.com/actions/setup-node/releases/tag/v4.0.0
+        uses: actions/setup-node@8f152de45cc393bb48ce5d89d36b731f54556e65
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm install
```
pedroferreira1, line 65: We should use `npm ci` for CI runners intead of `npm install`

tuliomir: Fixed on [50ca249](https://github.com/HathorNetwork/hathor-wallet-headless/pull/364/commits/50ca249687a0a8f0618b82110084f4ea8ce59c59)

pedroferreira1 (approved):

.github/workflows/apidocs-upload.yml
luislhl: I think it's good

luislhl (approved):

pedroferreira1 (approved):

PR #368, opened 2023-12-15
chore: production server without npm

r4mmer:
### Acceptance Criteria

- Do not use `npm` on docker entrypoint
- Do not use `babel-node` on production images
- Ensure wallet version is always using package.json version

### Explanation

During the QA of the release-candidate version we found an issue with npm running on read only filesystems.
Running production wallets with readonly fs is a security concern that cannot be ignored, so @luislhl investigated why this is happening and created a temporary fix that allowed npm to write its cache to a separate volume (in this [PR](https://github.com/HathorNetwork/ops-tools/pull/770)).

From the npm issues we find that this is a isolated issue when running npm and by running more recent images we find that this has not been fixed as of npm v10.

The solution is to stop using npm on our docker entrypoint and start using pure node.
When moving the npm start command to the entrypoint I found out that it uses babel-node and babel-node itself does not recommend using it on production environments (https://babeljs.io/docs/babel-node#not-meant-for-production-use).
So I moved to making the babel build during image build and running pure nodejs interpreter.

Some other best practices were incorporated, simply because they were easy to implement, and the more laborious ones will be made on a separate PR.

### Security Checklist


Dockerfile
```diff
@@ -3,12 +3,15 @@ WORKDIR /usr/src/app/
 
 COPY .babelrc package.json package-lock.json ./
 RUN apk add --no-cache --virtual .gyp python3 make g++ &&\
-    npm install &&\
+    apk add dumb-init &&\
```
r4mmer, line 6: dumb-init was added because nodejs is not meant to run as PID 1.
For instance it does not respond to sigint (ctrl-c) signals to stop the container.
dumb-init is a simple process supervisor and init system designed to run as PID 1 inside minimal container environments e.g. docker, which is perfect for our project

Dockerfile
```diff
@@ -3,12 +3,15 @@ WORKDIR /usr/src/app/
...
+    npm ci --only=production &&\
     apk del .gyp
 
 COPY ./src/ ./src/
 COPY ./scripts/ ./scripts/
 COPY config.js.docker ./src/config.js
 
+RUN npm run build
```
luislhl, line 14: suggestion: Maybe we could have a multi-stage dockerfile, because we only need the `dist` folder and `dumb-init` to run the program, right?

The rest is only needed for the build (`python3`, `make`, `g++`)

luislhl (dismissed):

package.json
```diff
@@ -21,13 +21,14 @@
     "yargs": "^16.2.0"
   },
   "scripts": {
+    "build": "babel src -d dist",
+    "dev": "nodemon --exec babel-node src/index.js",
+    "start": "npm run build && node dist/index.js",
```
tuliomir, line 26: question: This is a very welcome change :star2:, as it will allow us to add debuggers and breakpoints easier when debugging the headless wallet.

However, is it possible for us to add `.map` files? Otherwise we'll be debugging and receiving stack traces from transpiled code only. This may make the development process slower.

r4mmer: Done, but building sourceMaps by configuring it on the .babelrc file was only introduced in [v7.19.3](https://github.com/babel/babel/releases/tag/v7.19.3) and we are using 7.13, we should upgrade the dependencies, but I don't think this PR should cover this.

To build sourceMaps on v7.13 we can only do it by explicitly passing the `--source-maps` argument to babel cli unless we want the sourceMaps to be inline on the file but this option is not recommended for production environments.

Dockerfile
r4mmer: I think we should do this, but in another PR where we can include some other best practices and updates.

src/controllers/index.controller.js
```diff
@@ -12,7 +12,8 @@ const { notificationBus } = require('../services/notification.service');
 const { cantSendTxErrorMessage, API_ERROR_CODES } = require('../helpers/constants');
 const { parametersValidation } = require('../helpers/validations.helper');
 const { sanitizeLogInput } = require('../logger');
-const { getReadonlyWalletConfig, getWalletConfigFromSeed, WalletStartError } = require('../helpers/wallet.helper');
+const { getReadonlyWalletConfig, getWalletConfigFromSeed } = require('../helpers/wallet.helper');
+const { WalletStartError } = require('../errors');
```
tuliomir, line 16: nitpick: I'm curious: was this import incorrect?

tuliomir (approved):

src/controllers/index.controller.js
r4mmer: Yeah, this file does not export `WalletStartError`

luislhl (approved):

PR #373, opened 2023-12-19
feat: add support for multiple data outputs in the create token API

pedroferreira1:
### Acceptance Criteria
- Add support for multiple data outputs in the create token API

### Security Checklist

### DIFF
--- a/__tests__/integration/create-token.test.js
+++ b/__tests__/integration/create-token.test.js
@@ -484,4 +484,40 @@ describe('create token', () => {
 
     done();
   });
+
+  it('create token with data outputs', async done => {
+    const response = await TestUtils.request
+      .post('/wallet/create-token')
+      .send({
+        name: 'Token C',
+        symbol: 'TKC',
+        amount: 100,
+        data: ['test1', 'test2']
+      })
+      .set({ 'x-wallet-id': wallet2.walletId });
+
+    const transaction = response.body;
+    expect(transaction.success).toBe(true);
+
+    await TestUtils.waitForTxReceived(wallet2.walletId, response.body.hash);
+
+    // Make sure the token was minted
+    const tkBalance = await wallet2.getBalance(transaction.hash);
+    expect(tkBalance.available).toBe(100); // The newly minted tokens
+
+    // Make sure the last 2 outputs are the data outputs
+    const lastOutput = transaction.outputs[transaction.outputs.length - 1];
+    expect(lastOutput.value).toBe(1);
+    expect(lastOutput.tokenData).toBe(0);
+    const lastOutputScript = scriptsUtils.parseScriptData(Buffer.from(lastOutput.script.data));
+    expect(lastOutputScript.data).toBe('test2');
+
+    const outputBeforeLast = transaction.outputs[transaction.outputs.length - 2];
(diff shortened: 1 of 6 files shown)


src/api-docs.js
```diff
@@ -2600,6 +2600,13 @@ const defaultApiDocs = {
                     type: 'boolean',
                     description: 'If the melt authority address is allowed to be from another wallet. Default is false.'
                   },
+                  data: {
+                    type: 'array',
+                    items: {
+                      type: 'string'
+                    },
+                    description: 'List of strings to create a data output for each.'
```
r4mmer, line 2608:
```suggestion
                    description: 'List of utf-8 encoded strings to create a data output for each.'
```

__tests__/integration/create-token.test.js
```diff
@@ -484,4 +484,42 @@ describe('create token', () => {
...
+    const transaction = response.body;
+    expect(transaction.success).toBe(true);
+
+    await TestUtils.waitForTxReceived(wallet2.walletId, response.body.hash);
+
+    // Make sure the token was minted
+    const tkBalance = await wallet2.getBalance(transaction.hash);
+    expect(tkBalance.available).toBe(100); // The newly minted tokens
+
+    // Make sure the last 2 outputs are the data outputs
```
r4mmer, line 508: The ordering is guaranteed?
I think that the order can be random if there is a change output

__tests__/integration/create-token.test.js
```diff
@@ -484,4 +484,42 @@ describe('create token', () => {
...
+
+    const outputBeforeLast = transaction.outputs[transaction.outputs.length - 2];
+    expect(outputBeforeLast.value).toBe(1);
+    expect(outputBeforeLast.tokenData).toBe(0);
+    const outputBeforeLastScript = scriptsUtils.parseScriptData(
+      Buffer.from(outputBeforeLast.script.data)
+    );
+    expect(outputBeforeLastScript.data).toBe('test1');
+
+    done();
```
tuliomir, line 523: suggestion: Do you think it would be useful to add a validation that 3 HTR were spent in total to create this token?

When we test for the output values we know this spent exists from the perspective of the blockchain, but not from the perspective of the headless endpoints.

__tests__/integration/create-token.test.js
pedroferreira1: Actually it's not random. We shuffle the outputs only when we use the `sendManyOutputsTransaction`. All the others create the transaction object and just mine/push, so we don't shuffle the outputs.

__tests__/integration/create-token.test.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/373/commits/cb123251019073e716f1b578296561a7c36f83ff

src/api-docs.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-headless/pull/373/commits/4b964157a83e6f6d740bfcf9f977021ffd181310

r4mmer (approved):

tuliomir (approved):

PR #374, opened 2023-12-22
feat: multistage dockerfile

r4mmer:
### Acceptance Criteria

- Use a multistage dockerfile
- Move babel to dev dependency
- Use production node environment
- lock wallet-lib version

### Security Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ FROM $IMG as builder
 WORKDIR /usr/src/app/
 
 COPY .babelrc package.json package-lock.json ./
+# Install build dependencies
+# We use `npm install` so we install dev deps e.g. babel
 RUN apk add --no-cache --virtual .gyp python3 make g++ &&\
     npm install &&\
     apk del .gyp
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,17 +1,45 @@
...
+
+FROM $IMG as builder
+
 WORKDIR /usr/src/app/
 
 COPY .babelrc package.json package-lock.json ./
 RUN apk add --no-cache --virtual .gyp python3 make g++ &&\
-    apk add dumb-init &&\
-    npm ci --only=production &&\
+    npm install &&\
```
tuliomir, line 9: I understand why we're using the `install` command instead of `ci`, but it's unusual to see this in a production environment.

What do you think about adding a comment here just expliciting that the `install` is needed for the `builder` environment?

tuliomir (dismissed):

luislhl (approved):

Dockerfile
r4mmer: Done

tuliomir (approved):

PR #376, opened 2024-01-03
refactor: use new hathor-healthcheck-lib

luislhl:
### Acceptance Criteria
We should use the new hathor-healthcheck-lib to build the response of the healthcheck endpoint.

The lib is not yet releases on `npm`, so this PR is still waiting for it to be released.
However, the logic should be ready.

In case you want to test it locally, you'll have to build and pack the new lib and install it from your local package.

### TODO
- [x] Use the lib from npm

### Security Checklist


src/services/healthcheck.service.js
```diff
@@ -1,42 +1,96 @@
...
 
-const { buildComponentHealthCheck } = require('../helpers/healthcheck.helper');
 const { friendlyWalletState } = require('../helpers/constants');
 
-const healthService = {
+class HealthService {
+  constructor(walletIds, includeFullnode, includeTxMiningService) {
+    this.healthcheck = new Healthcheck({
+      name: "hathor-wallet-headless",
+      warn_is_unhealthy: true
```
tuliomir, line 12: question: Should this property be configurable?

tuliomir (commented):
:heavy_check_mark: This looks good to me. I'm only refraining from approving the PR because the `package.json` is yet to be adjusted with the final `npm` package name.

src/services/healthcheck.service.js
luislhl: I think so. We support 3 possible statuses for healthchecks, which are `pass`, `fail` and `warn`.

However, we need to translate them into a 200 or 503 HTTP status code, because most infrastructure systems only consider the status code for healthchecks, not the response body.

So, some projects using this lib could need to consider `warn` as a 503, and some could prefer to consider it a 200 (in case they have some other way to parse the body, for instance)

What do you think?

tuliomir: According to [the RFC we used as reference](https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check) for this, I understand the status for a `warn` in case we explicitly set it _not to be unhealthy_ should be `200`.

If a client application expects to receive a `503` on healthcheck warnings, the configuration for this headless instance should just be left as default - as it is right now with `warn_is_unhealthy: true`.

I can imagine some more complex use cases, but I think this boolean configuration with a default to `true` would suffice for almost all users.

Do you agree?

luislhl: Yes, I do. Done in https://github.com/HathorNetwork/hathor-wallet-headless/pull/376/commits/971d3b2df4220e04cfb9bed1b69b28a76f9f08ad

src/services/healthcheck.service.js
```diff
@@ -1,80 +1,134 @@
...
+import { Healthcheck, HealthcheckInternalComponent, HealthcheckHTTPComponent, HealthcheckCallbackResponse, HealthcheckStatus } from 'hathor-healthcheck-lib';
+import { initializedWallets } from './wallets.service';
+import { getConfig } from '../settings';
 
-const { buildComponentHealthCheck } = require('../helpers/healthcheck.helper');
 const { friendlyWalletState } = require('../helpers/constants');
 
-const healthService = {
+class HealthService {
+  constructor(walletIds, includeFullnode, includeTxMiningService) {
```
tuliomir, line 9: note(non-blocking): This constructor is making a lot of work behind the scenes and calling other functions.

As of now, all procedures are synchronous, but should any of them be refactored to `async`, we would need to change the way we build the HealthService.

Anticipating this possibility, I leave [this SO answer](https://stackoverflow.com/a/43433773/1123007) that has helped me before on a refactoring of this kind, mentioning patterns that involve `init()` class methods and `create()` static methods.

src/settings.js
```diff
@@ -62,6 +62,8 @@ const { initHathorLib } = require('./helpers/wallet.helper');
  * @property {string[]} [enabled_plugins] - The plugins we should start when
  * the service starts up.
  * @property {Record<string, PluginConfig>} [plugin_config] - The custom plugin configuration.
+ * @property {boolean} [considerHealthcheckWarnAsUnhealthy] - If true, the healthcheck
```
tuliomir, line 65: issue: This new property needs to be added to the `_analizeConfig()` processing as well. See [this change on PR 356](https://github.com/HathorNetwork/hathor-wallet-headless/pull/356/files#diff-f6437b5f2e4d199e7a66f03c9d33dc5e06f5d2edc29abc401417288a43633793) for reference.

This is so that the user can update its value without the need to restart the headless application.

tuliomir (changes requested):

src/services/healthcheck.service.js
luislhl: Thanks for the reference. I think needing to turn the methods into `async` is a remote possibility.

Even if we do need, we could just move them to an `async init()` and call it from the `healthcheck.controller.js`, which is already `async` as well.

But since you're considering this as non-blocking, I'll leave this refactor for when we actually need it.

src/settings.js
luislhl: I don't see why I would have to necessarily add it to `_analizeConfig()`, since this property shouldn't trigger any of the special cases treated there (stopAllWallets, reconfigLib and nonRecoverable).

It seems that a change to the property in the config would work the way it is implemented now.

If I'm missing something, I would need additional details from you.

tuliomir (approved):

src/settings.js
tuliomir: You're right: each call to the `healthcheck` endpoint will fetch the most updated configuration already.
Closing this thread.

src/services/healthcheck.service.js
```diff
@@ -1,80 +1,134 @@
...
    *
-   * @returns {Object}
+   * @returns {HealthcheckCallbackResponse}
    */
-  async getFullnodeHealth() {
+  static async getFullnodeHealth() {
     let output;
     let healthStatus;
 
     // TODO: We will need to parse the healthData to get the status,
```
pedroferreira1, line 102: Is this TODO still valid?

pedroferreira1 (approved):

src/services/healthcheck.service.js
luislhl: Yes, but I prefer to do this in a separate PR. We have a task to not forget about it.

pedroferreira1 (approved):

tuliomir (approved):

PR #399, opened 2024-02-08
chore: use hathor-core health endpoint

luislhl:
Depends on: https://github.com/HathorNetwork/hathor-wallet-lib/pull/608

### Acceptance Criteria
- We should consult hathor-core's health endpoint to build our own health response.

### Security Checklist

### DIFF
--- a/__tests__/test-utils.js
+++ b/__tests__/test-utils.js
@@ -198,7 +198,8 @@ class TestUtils {
 
   static startMocks() {
     // http mocks
-    httpMock.onGet('/version').reply(200, httpFixtures['/v1a/version']);
+    httpMock.onGet('http://fakehost:8083/v1a/version').reply(200, httpFixtures['http://fakehost:8083/v1a/version']);
+    httpMock.onGet('http://fakehost:8083/v1a/health').reply(200, httpFixtures['http://fakehost:8083/v1a/health']);
     httpMock
       .onGet('/thin_wallet/address_history')
       .reply(200, httpFixtures['/thin_wallet/address_history']);
--- a/src/services/healthcheck.service.js
+++ b/src/services/healthcheck.service.js
@@ -99,8 +99,6 @@ class HealthService {
     let output;
     let healthStatus;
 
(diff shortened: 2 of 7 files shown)


src/services/healthcheck.service.js
```diff
@@ -99,14 +99,21 @@ class HealthService {
...
-      healthStatus = HealthcheckStatus.PASS;
+      healthStatus = healthData.status;
+
+      const config = getConfig();
+
+      const isUnhealthy = healthStatus === HealthcheckStatus.FAIL
+        || (healthStatus === HealthcheckStatus.WARN && config.considerHealthcheckWarnAsUnhealthy);
+
+      if (isUnhealthy) {
+        output = `Fullnode reported as unhealthy: ${JSON.stringify(healthData)}`;
```
tuliomir, line 113: chore: Since we're adding complexity because of the increased possible results of the healthcheck, I think we should also increase the coverage here.

Codecov is informing that this line was never reached by tests. Could we add it to the test suite?

tuliomir (approved):

__tests__/test-utils.js
```diff
@@ -198,7 +198,8 @@ class TestUtils {
 
   static startMocks() {
     // http mocks
-    httpMock.onGet('/version').reply(200, httpFixtures['/v1a/version']);
```
pedroferreira1, line 201: Why did you have to change this?

pedroferreira1 (approved):

__tests__/test-utils.js
luislhl: I didn't have to. I just thought it would be clearer to add the full address.

The way these mocks are declared make it confusing to know whether these endpoints being mocked belong to wallet-headless itself, to the fullnode or to tx-mining-service.

Also to keep the same pattern as I used for the `/health` mock on the fullnode. For it I necessarily had to add the fullnode address, because the wallet-headless also has a `/health` endpoint.

We could undo the change for `/version`, though. Tests will keep working.

PR #400, opened 2024-02-09
docs(README): update and organize readme and supplementary documentation of the repository

BrunoCampana:
### Motivation

- README was outdated.
- Overall documentation of this repository was outdated.
- Documentation of codebase isn't standardized.

### Acceptance Criteria

- Update README.
- Organize the documentation of repository.
- Update non-default docs to standardize them.

Standards that must be followed: (all from `engineering-guidelines` repository)
- [README standard](https://github.com/HathorNetwork/engineering-guidelines/blob/master/documentation/write-readme-doc.md)
- [README template](https://github.com/HathorNetwork/engineering-guidelines/blob/master/documentation/readme-template-sw-component-repo.md)
- [Non-default docs standard](https://github.com/HathorNetwork/engineering-guidelines/blob/master/metadocs/drafts/write-ad-hoc-doc.md)
- [Organization of documentation](https://github.com/HathorNetwork/engineering-guidelines/blob/master/metadocs/drafts/high-level-view.md)

### Checklist

- [X] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged


docs/get-started.md
```diff
@@ -0,0 +1,332 @@
+# Get started
```
pedroferreira1, line 1: Why do we need this file? We don't have any link to it in the `README`.

README.md
```diff
@@ -1,344 +1,41 @@
...
-Check out how to configure and use as a MultiSig wallet on [MULTISIG.md](./MULTISIG.md)
+- [Testing guidelines](/docs/testing.md)
+- [QA: suggested test sequence](/docs/QA.md)
 
-## Testing
+## Contributing
 
-This application has two different contexts for testing.
-Refer to [TESTING.md](./TESTING.md) for more information on this.
+To contribute to the development of Hathor headless wallet, we encourage you to fork the `master` branch, implement your code, and then [open a pull request to merge it into `master`, selecting the "feature branch template"](https://github.com/HathorNetwork/hathor-wallet-headless/pulls).
```
pedroferreira1, line 32: I think we could remove this link, it just goes to the list of open pull requests. What do you think?

pedroferreira1 (commented):
I think we could have the API docs link somewhere in the README. Even though we have the link for the official Hathor docs and the API docs link there, I think we should have it in the README as well. 

What do you think?

docs/get-started.md
BrunoCampana: ### About the `get-started.md` file:

Change made: file deleted.
- About the `get-started.md` file, we actually don't need it.
- I moved the root headless miscellaneous documents from the repository to the `docs` directory.
- Then I indexed them all either in the main `README`, in the root of the repository or in the secondary, the `README` in the `docs` directory.
- This file was in fact not indexed because it is not relevant, as all of its content is already covered elsewhere.
- I should have deleted it after reviewing the contents of all the files, but I ended up forgetting.

README.md
BrunoCampana: Change made:
- The link was wrong. The link was pointing to the list of PRs, when it should actually point to the new PR opening page.
- Regarding the usefulness and relevance of this link, I agree that even after the change, it is not an indispensable link.
- We could remove it, but I think it helps to serve as a "call to action", in a discreet but relevant way for us to point out that we encourage and suggest collaborations from the entire community.

docs/get-started.md
BrunoCampana: **I think we shouldn't add the link to the API docs in the README.**
I'll explain the reasons I see for this:
- It's important that there is a single place where the reader can find everything they need to know about the software component in question.
- This place doesn't need to contain all the information, but should point out where to get all the information.
- For components covered by Hathor docs, this place should be the Hathor docs.
- For the case of components not covered by Hathor docs, this place should be the repository's own README.
- To note: today only two components are covered in Hathor docs: core and headless.
- However, Hathor docs lists all the main components of the Hathor system, and as it doesn't document, for example, the desktop wallet, it provides the link to its repository.
- My effort has been towards REDUCING information duplication, and making the user journey clearer.
- (1) They want to take a look at the Docker images we have? Go to Docker Hub.
- (2) They need support? Go to Discord.
- (3) They need technical documentation for operation and use? Hathor docs, and if not there, they will be redirected to the README of the respective repo.
- (4) They want to see the source code, want to open an issue, want to collaborate, etc. They go to the Github repo.
- For example, until recently, we had installation instructions in the repo's README, on Gitbook, and in the Docker Hub images repo's README. This is not good: it's bad for us, because it makes it more difficult to keep everything updated, and it's bad for the reader, because it leaves them wondering "which installation guide should I follow", the one on the Github repo, the one on the Docker Hub repo or the one on Gitbook?
- Because of all this, my proposal for standardization is as follows: the main point of technical documentation for operation and use of the main SW components of our codebase should be the Hathor docs website. Our commitment should be to provide guidance to other places from there.
- We need to make sure the reader has no doubt: everything they need will be on Hathor docs. If the content is not there, at least the link to where they will find what they need should be there.
- That said: it may seem harmless to have a direct link to the API docs in the README. However, think about the following: the entry point for documentation for many users will not be the documentation website. Out of habit, they will search for the repository on Github. But from there, we have to direct them to the place where we guarantee they will find everything they need, that is, Hathor docs.
- If we give a direct link for them to access the API docs from Github, they go and read the API documentation. And they might think that's all the documentation we have to offer on using the API and keep using just that, when in fact, on Hathor docs, we have written several articles to better explain the use of the APIs together.
- I'll give an example: if they want to know how to implement atomic swaps.
- Then they go to the Github repo. Then, instead of going through Hathor docs, they go straight to the API docs. Then they think: "well, here are the APIs described correctly, but I don't know how to use all these API operations together". Then they will think that's all the documentation there is for using the APIs, when in fact there's much more.
- If they access the main page of headless on Hathor docs, they will see all the resources we have available for them to operate and use that software.

**Beyond all this: I'm working on a side project to add the API documentation, just like redoc does, but within our documentation website. Thus, soon the API documentation will not be on a separate website, but within the main website.**

pedroferreira1 (approved):

msbrogli (dismissed):

pedroferreira1 (approved):

msbrogli (approved):

PR #405, opened 2024-02-16
chore: Upgrades `Node.js`

tuliomir:
### Acceptance Criteria
- Upgrades `Node.js` to the latest LTS: v20

### Notes
A local docker instance was built with this code and performed successfully on the same operations executed by the `wallets-monitor`

It will also be necessary to change the necessary CI steps for this repository. Currently the tests on version `14` and `16` are mandatory for merging the PR.

### Security Checklist


.github/workflows/integration-test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [20.x]
```
tuliomir, line 27: Here we could have tests running on `[18.x, 20.x]`, but at first I see no benefit in this extra round of tests on a version that's already in maintenance mode.

luislhl: Also, in case people always use the Docker image to run the wallet, there is a guarantee that we only need to support a specific version.

README.md
```diff
@@ -7,7 +7,7 @@ It may be used both by mining pools and exchanges.
 
 ## Requirements
 
-* [Node.js](https://nodejs.org/): We recommend and test with version >= 14.0.0, but version >= 10.0.0 will also probably work. Do not use versions < 10.0.0, they are known to have issues.
+* [Node.js](https://nodejs.org/): We recommend and test with version >= 20.0.0.
```
luislhl, line 10: suggestion: People won't be able to use any lower version anyway, since we have `engine-strict=true` in the `.npmrc` file.

Should we change this sentence to reflect this?

luislhl (approved):

pedroferreira1 (approved):

README.md
tuliomir: Done on [eca6313](https://github.com/HathorNetwork/hathor-wallet-headless/pull/405/commits/eca6313bace5fed042cf70a10ffbc2dc571e1224)

Issue #269, opened 2023-02-01
Create integration tests for provided plugins


r4mmer:
Create a suite of integration tests to run the plugins and check that the events match the expected events

PR #407, opened 2024-02-28
test: Plugin events integration tests

tuliomir:
This PR adds integration tests for the Plugins, validating their event handling behavior on a live blockchain.

The initial expectation here, as described on #269,  was to test the actual plugin files packed with the Headless wallet. However, that would require new test workflows, with their configuration files and dependencies to be added, thus increasing the complexity of the tests suite.

A decision was made to restrict the scope to only testing the event bus on the integration environment: this validates that all events are sent in the correct moment with the correct data. An exception is made to the _New Best Blockchain_ event that is still untested because it requires a more complex setup.

This covers the gap of having integration tests for the plugins and only requires one test file and one custom plugin, with no need to change any infrastructure or dependency. The remaining behaviors related to command line parameters and dependency method calls for each plugin are being tested by the unit tests already.

### Acceptance Criteria
- Should validate the event handling of the plugins on integration tests

### Security Checklist

### DIFF
--- a/__tests__/integration/plugins/plugin-events.test.js
+++ b/__tests__/integration/plugins/plugin-events.test.js
@@ -11,17 +11,23 @@ describe('test the plugin event handler', () => {
   let wallet1;
 
   beforeAll(async () => {
-    // Mimic the behavior or `src/plugins/child.js`, that is not called within the context of Jest
+    /*
+     * Mimic the behavior of `src/plugins/child.js` because it is not called within the context of
+     * Jest. However, here we initialize the plugin within the main process to allow for integration
+     * testing.
+     *
+     * @see https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/src/index.js
+     * @see https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/src/plugins/child.js
+     */
     childManager.hathorPlugins.integration = {
       name: 'integration',
       file: '../../__tests__/integration/plugins/integration_test_plugin',
     };
-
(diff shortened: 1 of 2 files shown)


__tests__/integration/plugins/plugin-events.test.js
```diff
@@ -0,0 +1,92 @@
...
+  it('should return the events that led to complete wallet start', async () => {
+    const currentHistory = loadedPlugin
+      .retrieveEventHistory();
+
+    // The full event history should have more elements than those of the test wallet
+    const testWalletHistory = currentHistory.filter(event => event.walletId === pluginWalletId);
+    expect(currentHistory.length).toBeGreaterThan(testWalletHistory.length);
+
+    // There must have been 5 wallet changes for this test, one with each step of its initialization
+    expect(testWalletHistory.filter(event => event.type === 'wallet:state-change')).toHaveLength(5);
```
r4mmer, line 52: suggestion(if-minor): test that we have `wallet:state-change` instead of the number of state changes.

This is just to avoid future changes on the lib to have an effect on these tests.

r4mmer (approved):

__tests__/integration/plugins/plugin-events.test.js
tuliomir: Good point. Done on [182dbec](https://github.com/HathorNetwork/hathor-wallet-headless/pull/407/commits/182dbec7a0d4f11cf811a29b5e2f7dc3edf33a02).

r4mmer (approved):

pedroferreira1 (approved):

PR #410, opened 2024-02-29
feat: hsm integration (phase 2)

r4mmer:
### Acceptance Criteria

- Allow signing and pushing transactions with HSM

### Security Checklist


src/services/hsm.service.js
```diff
@@ -6,16 +6,226 @@
...
+        signature,
+        pubkey,
+      );
+      tx.inputs[index].setData(inputData);
+    }
+    return tx;
+  }
+}
+
+async function hsmStartSession(rootKeyname) {
```
andreabadesso, line 221: Missing docstring

src/services/hsm.service.js
```diff
@@ -98,40 +308,50 @@ async function isKeyValidXpriv(hsmConnection, hsmKeyName) {
...
+ * @returns {Promise<boolean>}
+ */
+async function hsmKeyExists(hsmConnection, keyname) {
+  try {
+    await hsmConnection.blockchain.getKeyInfo(keyname);
+    return true;
+  } catch (e) {
+    // hsm.constants.ERR_CANNOT_OPEN_OBJ = 5004
+    if (e.errorCode === hsm.constants.ERR_CANNOT_OPEN_OBJ) {
+      await delay(500);
```
andreabadesso, line 323: Can you please add a comment explaining why we need this delay?

src/services/hsm.service.js
```diff
@@ -141,78 +361,57 @@ async function derivateHtrCkd(
...
-    hsm.enums.BCHAIN_SECURE_BIP32_INDEX.BASE,
-    false,
-    true,
-    childKeyNames.HTR_CKD_COIN_KEYNAME,
-    childKeyNames.HTR_CKD_ACCOUNT_KEYNAME,
-  );
+    // Derivation 2: Coin
+    await hsmConnection.blockchain.createBip32ChildKeyDerivation(
+      derivationVersion,
+      hsm.enums.BCHAIN_SECURE_BIP32_INDEX.BASE + 280,
```
andreabadesso, line 384: We have 280 as a constant on the lib, don't we?

src/services/hsm.service.js
```diff
@@ -6,16 +6,208 @@
...
+      dataToSignHash,
+      htrKeyName,
+    );
+
+    // Dinamo SDK returns v + DER, where v is a parity byte
+    // https://manual.dinamonetworks.io/nodejs/enums/hsm.enums.BLOCKCHAIN_SIG_TYPE.html#SIG_DER_ECDSA
+    // We can safely ignore the parity byte
+    if (signature[0] !== 0x30) {
+      return signature.slice(1);
+    }
```
andreabadesso, line 138: So what this is doing is ignoring every byte that is not a sequence byte, can you give more context on this in the comment below?

Also, can 0x30 be a constant?

src/services/hsm.service.js
```diff
@@ -141,78 +343,57 @@ async function derivateHtrCkd(
...
-    true, // Temporary
-    hsmKeyName, // Parent key name
-    childKeyNames.HTR_CKD_BIP_KEYNAME, // Child key name
-  );
+  if (!acctPathExists) {
+    // derive from root to account path
+    const childKeyNames = {
+      HTR_CKD_BIP_KEYNAME: 'HTR_BIP_KEY',
+      HTR_CKD_COIN_KEYNAME: 'HTR_COIN_KEY',
+    };
```
andreabadesso, line 351: Should this be in the constants file?

andreabadesso (changes requested):

src/services/hsm.service.js
```diff
@@ -6,16 +6,208 @@
...
+      }
+      const addressInfo = await storage.getAddressInfo(spentOut.decoded.address);
+      if (!addressInfo) {
+        // Not a wallet address
+        continue;
+      }
+
+      const pubkeyHex = await storage.getAddressPubkey(addressInfo.bip32AddressIndex);
+      const pubkey = Buffer.from(pubkeyHex, 'hex');
+      const signature = await this.signData(dataToSignHash, addressInfo.bip32AddressIndex);
```
andreabadesso, line 170:
```suggestion
      const [pubkeyHex, signature] = await Promise.all([
        storage.getAddressPubkey(addressInfo.bip32AddressIndex),
        this.signData(dataToSignHash, addressInfo.bip32AddressIndex)
      ]);
      const pubkey = Buffer.from(pubkeyHex, 'hex');
```

andreabadesso (changes requested):

src/controllers/hsm/hsm.controller.js
```diff
@@ -112,22 +123,159 @@ async function startHsmWallet(req, res) {
...
+  const validationResult = parametersValidation(req);
+  if (!validationResult.success) {
+    res.status(400).json(validationResult);
+    return;
+  }
+
+  if (!isHsmWallet(req.headers['x-wallet-id'])) {
+    res.status(400).json({ success: false, message: 'This endpoint can only be used with an HSM wallet' });
+    return;
+  }
```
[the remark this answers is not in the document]
r4mmer: This endpoint has been removed since we added support for HSM on all endpoints.
We can use the simple-send-tx endpoint as normal now.

src/controllers/hsm/hsm.controller.js
```diff
@@ -112,22 +123,159 @@ async function startHsmWallet(req, res) {
...
+}
+
+/**
+ * Send a single output transaction (plus change if needed)
+ *
+ * @param {Request} req
+ * @param {Response} res
+ */
+async function simpleSendTx(req, res) {
+  const validationResult = parametersValidation(req);
```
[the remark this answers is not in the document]
r4mmer: This has been removed

src/services/hsm.service.js
```diff
@@ -6,16 +6,208 @@
...
+      await this.conn.blockchain.createBip32ChildKeyDerivation(
+        hsm.enums.VERSION_OPTIONS.BIP32_HTR_MAIN_NET,
+        index,
+        false,
+        true,
+        changeKeyName,
+        addressKeyName,
+      );
+    } catch (err) {
+      await delay(500);
```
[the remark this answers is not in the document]
r4mmer: Added explanation

src/services/hsm.service.js
```diff
@@ -6,16 +6,208 @@
...
+   * @param {HsmSession} hsmSession
+   * @param {Buffer} dataToSignHash
+   * @param {number} addressIndex
+   * @returns {Promise<Buffer>} DER encoded signature
+   */
+  async signData(dataToSignHash, addressIndex) {
+    // Derive the key to the desired address index
+    const htrKeyName = await this.getAddressKeyName(addressIndex);
+
+    const signature = await this.conn.blockchain.sign(
```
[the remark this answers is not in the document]
r4mmer: Done

src/services/hsm.service.js
r4mmer: Not every byte, just the first byte if it is not `0x30`.
The `0x30` in DER encoding is a tag byte that marks the subsequent data as a sequence (or array).
Since a DER signature is the `[r,s]` array it should always start with `0x30`, but dinamo adds a parity byte.

We can use the parity byte to check the signature but it is not required.
Our lib (bitcore-lib) will throw an error if the first byte is not `0x30`, so we have to remove it either way.

src/services/hsm.service.js
r4mmer: These are temporary key names that are used only once during the first start of the wallet.
So I think its safe to have them here.

src/services/hsm.service.js
r4mmer: I understand the suggestion but the first promise is just a single derivation from the xpubkey and the other is an api call.
So I don't think we would save a lot of time with this approach.

src/services/hsm.service.js
r4mmer: Added docstring

src/services/hsm.service.js
r4mmer: Done

src/services/hsm.service.js
r4mmer: Changed to use this constant

src/services/hsm.service.js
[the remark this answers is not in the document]
andreabadesso: Can you add a TODO so we can grep todos in the future?

andreabadesso (approved):

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
 
 const { getConfig } = require('../settings');
 const { lock, lockTypes } = require('../lock');
 const { hsmBusyErrorMessage } = require('../helpers/constants');
 const { HsmError } = require('../errors');
 
+/* istanbul ignore next */
+/**
+ * After an error, the HSM lib has a runtime issue that will fail the next call
+ * To avoid this we can wait 500ms before continueing.
```
pedroferreira1, line 19:
```suggestion
 * To avoid this we can wait 500ms before continuing.
```

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+    name => `${name}_htrChange0`,
+    /**
+     * This unconventional naming is due to the 32 chars keyname restriction,
+     * we reserve 17 chars for context of which "*_HAddr_*" uses 7, leaving 10
+     * for the address index, this allows us to use all 2 billion indexes
+     * without resorting to hex conversion since 2 billion in base 10 has
+     * length 10.
+     */
+    (name, opts) => `${name}_HAddr_${opts.index}`,
+  ];
```
pedroferreira1, line 52: I didn't understand the need for this array. Why don't you have only this?

```
const levels = [
  keyname,
  `${keyname}_bip44`,
  `${keyname}_coinHTR`,
  `${keyname}_htrAcct`,
  `${keyname}_htrChange0`,
  `${keyname}_HAddr_${options.index}`
];
```

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+  async getAddressKeyName(index) {
+    if (index in this.addressKeys) {
+      return this.addressKeys[index];
+    }
+
+    const addressKeyName = getBip32Keyname(this.rootKey, 5, { index });
+    const changeKeyName = getChangeKeyname(this.rootKey);
+    try {
+      await this.conn.blockchain.createBip32ChildKeyDerivation(
+        hsm.enums.VERSION_OPTIONS.BIP32_HTR_MAIN_NET,
```
pedroferreira1, line 101: Can we have support for testnet also?

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+        index,
+        false,
+        true,
```
pedroferreira1, line 104: What are these parameters?

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+      throw err;
+    }
+
+    this.addressKeys[index] = addressKeyName;
+    return addressKeyName;
+  }
+
+  /**
+   * Sign the given data with the given HSM key
+   * @param {HsmSession} hsmSession
```
pedroferreira1, line 130: We don't have this parameter

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+      hsm.enums.BLOCKCHAIN_SIG_TYPE.SIG_DER_ECDSA,
+      hsm.enums.BLOCKCHAIN_HASH_MODE.SHA256,
+      dataToSignHash,
+      htrKeyName,
+    );
+
+    // Dinamo SDK returns v + DER, where v is a parity byte
+    // https://manual.dinamonetworks.io/nodejs/enums/hsm.enums.BLOCKCHAIN_SIG_TYPE.html#SIG_DER_ECDSA
+    // We can safely ignore the parity byte
+    if (signature[0] !== 0x30) {
```
pedroferreira1, line 152: I read an explanation about this value in another commend you replied to Abadesso, can you add it in the code?

> The 0x30 in DER encoding is a tag byte that marks the subsequent data as a sequence (or array). Since a DER signature is the [r,s] array it should always start with 0x30, but dinamo adds a parity byte.

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+  }
+
+  /**
+   * Sign the given data with the given HSM key
+   * @param {HsmSession} hsmSession
+   * @param {Buffer} dataToSignHash
+   * @param {number} addressIndex
+   * @returns {Promise<Buffer>} DER encoded signature
+   */
+  async signData(dataToSignHash, addressIndex) {
```
pedroferreira1, line 135: Maybe change the name to `getSignature`, since you are not really signing anything here. What do you think?

src/services/hsm.service.js
```diff
@@ -6,16 +6,210 @@
...
+    return signature;
+  }
+
+  /**
+   * Sign the given transaction and return the signatures.
+   * @param {hathorLib.Transaction} tx
+   * @param {hathorLib.Storage} storage
+   * @returns {Promise<SignatureData[]>}
+   */
+  async signTx(tx, storage) {
```
pedroferreira1, line 164: Don't we have a similar method in the lib? This seems a method exactly like any other `signTx` method, just with the custom `getSignature/signData`. If we don't have it, I think we should, this is an important method to have in the lib.

src/services/hsm.service.js
```diff
@@ -141,78 +348,57 @@ async function derivateHtrCkd(
...
-    return {
-      success: true,
-      htrKeyName: childKeyNames.HTR_CKD_ACCOUNT_KEYNAME,
-    };
+    // Derivation 3: Account
+    await hsmConnection.blockchain.createBip32ChildKeyDerivation(
+      derivationVersion,
+      hsm.enums.BCHAIN_SECURE_BIP32_INDEX.BASE,
+      false,
+      false,
```
pedroferreira1, line 383: Why this is `false` and in all the others this parameter is `true`?

src/services/hsm.service.js
```diff
@@ -141,78 +348,57 @@ async function derivateHtrCkd(
...
-    };
+  const changePathExists = await hsmKeyExists(hsmConnection, changePathKeyname);
+  if (!changePathExists) {
+    // derive from account path to change path 0
+    // Derivation 3: Account
+    await hsmConnection.blockchain.createBip32ChildKeyDerivation(
+      derivationVersion,
+      0,
+      false,
+      false,
```
pedroferreira1, line 397: Why `false`?

src/controllers/hsm/hsm.controller.js
```diff
@@ -94,10 +99,10 @@ async function startHsmWallet(req, res) {
   // Obtains this wallet's xPub from the HSM
   let xPub = null;
   try {
-    xPub = await hsmService.getXPubFromKey(connectionObj, hsmKeyName, {
-      isReadOnlyWallet: true,
-    });
+    await hsmService.deriveMainKeysFromRoot(connectionObj, hsmKeyName);
```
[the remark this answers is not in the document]
r4mmer: Not at all, but the method was generating all temporary keys.
After some dicussion with the devs at dinamo they recommended that some derivations are created as not temporary keys, the most important being the  account and change derivation.

The one before would derive to account and stop, this one derives to change making them persistent on the HSM.
This method also checks the existence of the keys before deriving (since they are persistent) saving time on subsequent initializations.

src/controllers/hsm/hsm.controller.js
```diff
@@ -109,23 +114,26 @@ async function startHsmWallet(req, res) {
 
   // Builds the wallet configuration object
   const walletConfig = getReadonlyWalletConfig({ xpub: xPub });
+  walletConfig.isSignedExternally = true;
```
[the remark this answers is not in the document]
r4mmer: Correct, removed

src/controllers/hsm/hsm.controller.js
```diff
@@ -109,23 +114,26 @@ async function startHsmWallet(req, res) {
...
+  const store = new hathorLib.MemoryStore();
+  const storage = new hathorLib.Storage(store);
+  // When signing transactions, the wallet will use this function
+  storage.setTxSignatureMethod(hsmService.hsmSignTxMethodBuilder(hsmKeyName));
```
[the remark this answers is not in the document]
r4mmer: Done

src/services/hsm.service.js
r4mmer: Done

src/services/hsm.service.js
r4mmer: This new levels array will be created when the method is called, all levels will be formatted.
The issue is the last one, `${keyname}_HAddr_${options.index}` will be an error on all derivations from level 0 to 4 since we are acessing `index` from the undefined `options`.

Making them functions will only format the one we are requesting.

src/services/hsm.service.js
r4mmer: Added comments explaining

src/services/hsm.service.js
r4mmer: Removed

src/services/hsm.service.js
r4mmer: This method is the equivalent of the `transactionUtils.getSignaturesForTx`, but instead of deriving the main key and signing it uses the dinamo api to do so.
It also uses the cached address keys to avoid deriving the same address twice during the signature process.

src/services/hsm.service.js
r4mmer: It is explained on the other comment

src/services/hsm.service.js
r4mmer: Done

src/services/hsm.service.js
r4mmer: Done

src/services/hsm.service.js
pedroferreira1: I think the method is really confusing as it is now. If the problem is the undefined, we can use this. What do you think?

```
const levels = [
  keyname,
  `${keyname}_bip44`,
  `${keyname}_coinHTR`,
  `${keyname}_htrAcct`,
  `${keyname}_htrChange0`,
];

if (options.index) {
  levels.push(`${keyname}_HAddr_${options.index}`);
}
```

src/services/hsm.service.js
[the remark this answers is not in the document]
r4mmer: Done

src/middlewares/wallet.middleware.js
```diff
@@ -51,6 +51,17 @@ async function walletMiddleware(req, res, next) {
   // Adding to req parameter, so we don't need to get it in all requests
   req.wallet = wallet;
   req.walletId = walletId;
+
+  /* istanbul ignore if  */
```
andreabadesso, line 55: Why is this needed?

andreabadesso (approved):

src/middlewares/wallet.middleware.js
r4mmer: It was meant for the if below, since it's unreachable in normal circunstances.
This was just to remove some lines from coverage that could not be tested.

src/middlewares/wallet.middleware.js
```diff
@@ -51,6 +51,16 @@ async function walletMiddleware(req, res, next) {
...
+
+  if (isHsmWallet(walletId)) {
+    const hsmKeyName = hsmWalletIds.get(walletId);
+    if (!hsmKeyName) {
+      sendError('Invalid HSM wallet.');
+      return;
+    }
+    req.hsmKeyName = hsmKeyName;
+  }
+
```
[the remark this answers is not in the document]
r4mmer: The `walletId` could be null or empty, but I think we can also test this on the `isHsmWallet` to make the middleware simpler

src/services/hsm.service.js
r4mmer: Done

pedroferreira1 (approved):

andreabadesso (approved):

PR #430, opened 2024-04-12
fix: run scripts in container

r4mmer:
### Acceptance Criteria
- Move utility commands for the user to Makefile
- Cleanup unused makefile commands
- Run commands with makefile

### Security Checklist


Makefile
```diff
@@ -1,27 +1,68 @@
 
+ifndef VERBOSE
+.SILENT:
+endif
```
r4mmer, line 4: This will block the stdout from make, so the output will be only from the script (unless we add the `VERBOSE` flag)

Makefile
```diff
@@ -1,27 +1,68 @@
...
+clean: script-clean-dirs
+
+.PHONY: build
+build: script-build-dirs
 
-docker_subtag := dev-$(shell git describe --tags --dirty)
-docker_tag := hathor-wallet-service:$(docker_subtag)
-docker_build_arg :=
-docker_build_flags :=
-ifneq ($(docker_build_arg),)
```
r4mmer, line 12: Docker is not built with the makefile anymore, so I'm removing these commands

Makefile
```diff
@@ -1,27 +1,68 @@
...
-	docker_build_flags +=  --build-arg $(docker_build_arg)
+.PHONY: script-build-dirs
+script-build-dirs:
+ifneq ("$(HEADLESS_SCRIPTS_SKIP_BUILD)", "1")
+	# Building scripts...
+	if [ -d "./dist" ]; then rm -rf ./dist; fi
+	if [ -d "./dist-scripts" ]; then rm -rf ./dist-scripts; fi
+	npm run --silent build > '/dev/null' 2>&1
+	npm run --silent build-scripts > '/dev/null' 2>&1
 endif
```
r4mmer, line 28: This command removes the current build and builds the source code, but only if `HEADLESS_SCRIPTS_SKIP_BUILD` is **not** `"1"`, since this is defined on the Dockerfile the container will not run this.

tuliomir (dismissed):
suggestion: It's good to have explanations on the PR comments for the `Makefile` clauses, but I think it would be important to have those explanations on the code as well.

That way, future maintenance would be facilitated with a clearer understanding of the scripts objectives, instead of having to refer to the git history and related PRs.

package.json
```diff
@@ -30,10 +30,7 @@
     "test_network_down": "docker-compose -f ./__tests__/integration/docker-compose.yml down && rm ./tmp/wallets.json",
     "lint": "eslint .",
     "lint_fix": "eslint . --fix",
-    "generate_words": "babel-node scripts/generate_words.js",
     "generate_wallets": "babel-node scripts/generate_wallets.js",
```
pedroferreira1, line 33: Why didn't you remove this?

pedroferreira1 (approved):

package.json
r4mmer: Generate wallets may be a script but it is intended to generate pre-calculated wallets for the integration tests.
The only commands moved to the Makefile were the ones intended to be called by the user.

Makefile
```diff
@@ -1,27 +1,68 @@
...
-ifneq ($(docker_build_arg),)
-	docker_build_flags +=  --build-arg $(docker_build_arg)
+.PHONY: script-build-dirs
+script-build-dirs:
+ifneq ("$(HEADLESS_SCRIPTS_SKIP_BUILD)", "1")
+	# Building scripts...
+	if [ -d "./dist" ]; then rm -rf ./dist; fi
+	if [ -d "./dist-scripts" ]; then rm -rf ./dist-scripts; fi
+	npm run --silent build > '/dev/null' 2>&1
+	npm run --silent build-scripts > '/dev/null' 2>&1
```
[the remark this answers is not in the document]
r4mmer: Done

pedroferreira1 (approved):

tuliomir (approved):

PR #439, opened 2024-05-11
chore: configure fireblocks secret in docker

r4mmer:
### Description

The fireblocks secret is a PEM encoded private key, which means it is a multiline file.
Since our config is an object I encoded the file contents in a single line with `\n` separators where the line breaks should be.
This approach works for the config file (i.e. running locally on the repo), but it did not work consistently in docker.

Docker would escape the `\n` in the string making them `\\n` and the content unusable.
There is a way to use multiline environment variable but it only works on older versions of docker.

To solve this I decided to actually use a file when running on docker and the config would be the file path.
Since the file needs to be in the container it can be mounted as a volume, which works when running with docker, docker-compose and Kubernetes.

The example command would be:

```bash
docker run --rm \
  -p 8000:8000 \
  -v ${PWD}/fireblocks.secret:/sec/fireblocks.secret \
  -e HEADLESS_SEED_DEFAULT='default'
  -e HEADLESS_FIREBLOCKS_API_KEY='123' \
  -e HEADLESS_FIREBLOCKS_API_SECRET_FILE='/sec/fireblocks.secret' \  # <--- This is the config being affected
  -e HEADLESS_NETWORK=testnet \
  -e HEADLESS_SERVER=https://node1.testnet.hathor.network/v1a/ \
  hathornetwork/hathor-wallet-headless
```

### Acceptance Criteria

- When running on docker make `fireblocks_api_secret` config be a filepath which will read the file contents into the actual config.

### Security Checklist

### DIFF
--- a/config.js.docker
+++ b/config.js.docker
@@ -1,3 +1,4 @@
+import 'fs';
 import yargs from 'yargs/yargs';
 import { hideBin } from 'yargs/helpers';
 
@@ -65,7 +66,6 @@ const config = {
 
   fireblocksUrl: getParam('fireblocks_url', 'https://api.fireblocks.io'),
   fireblocksApiKey: getParam('fireblocks_api_key'),
-  fireblocksApiSecret: getParam('fireblocks_api_secret'),
 
   http_api_key: getParam('api_key'),
   httpLogFormat: getParam('http_log_format'),
@@ -140,6 +140,11 @@ if (debugPath) {
     filename: debugPath,
     level: 'debug',
   };
+}
+
+const fireblocksApiSecretFile = getParam('fireblocks_api_secret');
+if (fireblocksApiSecretFile) {
+  config.fireblocksApiSecret = fs.readFileSync(fireblocksApiSecretFile);
 }
 
 module.exports = config;
--- a/config.js.docker
+++ b/config.js.docker
@@ -1,4 +1,3 @@
-import fs from 'fs';
 import yargs from 'yargs/yargs';
 import { hideBin } from 'yargs/helpers';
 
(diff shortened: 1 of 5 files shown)


tuliomir (approved):
We should just update the PR description to indicate an exception will be thrown if the configuration is incorrect for the Docker execution.

config.js.template
```diff
@@ -103,3 +109,9 @@ module.exports = {
   // Check the /health endpoint documentation for more info.
   considerHealthcheckWarnAsUnhealthy: true,
 };
+
+if (config.fireblocksApiSecretFile && fs.existsSync(config.fireblocksApiSecretFile)) {
+  config.fireblocksApiSecret = fs.readFileSync(config.fireblocksApiSecretFile, { encoding: 'utf8' }).trim();
+}
+
```
pedroferreira1, line 116: Shouldn't we do this in the `/start` controller? It's better than having this code in the config file.

pedroferreira1: This would remove some code from the docker config file as well

config.js.template
```diff
@@ -26,9 +28,13 @@ module.exports = {
   hsmUsername: null,
   hsmPassword: null,
 
+  // Only change the URL if you want to use the Sandbox API under 'https://sandbox-api.fireblocks.io'
   fireblocksUrl: 'https://api.fireblocks.io',
   fireblocksApiKey: null,
+  // This should be a string containing the secret (with line breaks)
```
pedroferreira1, line 34: Is it possible to use multiline strings here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#multi-line_strings

r4mmer: Yes, I personally used multiline string when developing and it worked without any issue.

config.js.template
r4mmer: Done

pedroferreira1 (approved):

config.js.docker
```diff
@@ -142,4 +142,15 @@ if (debugPath) {
   };
 }
 
+const fireblocksApiSecret = getParam('fireblocks_api_secret');
+if (fireblocksApiSecret) {
+  console.warn('[WARNING] fireblocks_api_secret is disabled for docker. Use fireblocks_api_secret_file instead.');
```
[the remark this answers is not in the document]
r4mmer: done

luislhl (approved):

PR #447, opened 2024-05-20
docs: fireblocks docs

r4mmer:
### Acceptance Criteria
- Document how to use the fireblocks integration

### Security Checklist

### DIFF
--- a/docs/fireblocks.md
+++ b/docs/fireblocks.md
@@ -1,0 +1,92 @@
+# Fireblocks integration
+
+## Description
+
+The Fireblocks integration allows you to connect your wallet with Fireblocks using the RAW transaction signing method.
+This means that the transaction is managed by the headless wallet and not by the Fireblocks API or dashboard.
+
+## Configuration
+
+The configuration requires an api key and secret which can be created on the Fireblocks dashboard.
+Under `Settings > Users` you can create a new API user with the `Editor` role and copy the API key and secret.
+
+The API key and secret are used to authenticate requests to the Fireblocks API.
+You will also need the account level xPub to start your wallet.
+
+### Docker
+
+When running on docker you need to add the secret as a file in the container, this can be used with [volumes](https://docs.docker.com/storage/volumes/).
+Then you need to configure the headless with the API key and path to the secret file in the container environment, you can use either cli arguments or environment variables to pass these configs.
+
+Environment variables:
+
+- `HEADLESS_FIREBLOCKS_API_KEY`: API key string
+- `HEADLESS_FIREBLOCKS_API_SECRET_FILE`: path to API secret file
+
+CLI arguments:
+
+- `--fireblocks-api-key`: API key string
+- `--fireblocks-api-secret-file`: path to API secret file
+
+### Running locally
+
+When running locally you write the `src/config.js` file and set the configuration in it.
+The config names are slightly different but work the same way as in the docker config.
+
+- `fireblocksApiKey`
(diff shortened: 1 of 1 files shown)


docs/fireblocks.md
```diff
@@ -0,0 +1,92 @@
+# Fireblocks integration
+
+## Description
+
+The Fireblocks integration allows you to connect your wallet with Fireblocks using the RAW transaction signing method.
```
tuliomir, line 5: suggestion: I see that this article is aimed at users that wish to use Fireblocks with the Headless Wallet.

However, with some changes it may also serve people are studying all possibilities of the Headless Wallet but do not yet know what Fireblocks is.

I believe that a brief explanation of Fireblocks with some explanatory links may help this other audience with little change.

docs/fireblocks.md
```diff
@@ -0,0 +1,92 @@
...
+This means that the transaction is managed by the headless wallet and not by the Fireblocks API or dashboard.
+
+## Configuration
+
+The configuration requires an api key and secret which can be created on the Fireblocks dashboard.
+Under `Settings > Users` you can create a new API user with the `Editor` role and copy the API key and secret.
```
tuliomir, line 11: suggestion: I'd recommend adding a link to one of their own tutorials about the Fireblocks service. Those navigation instructions can become obsolete and it can take a while for us to notice and to update this document.

docs/fireblocks.md
```diff
@@ -0,0 +1,92 @@
...
+
+Example:
+
+```bash
+curl -X POST -H "Content-Type: application/json" --data '{"xpub": "xpubABC...", "wallet-id": "w1"}' 'http://localhost:8000/fireblocks/start'
+```
+
+The `w1` wallet will now be connected to Fireblocks, you can use the usual endpoints to send transactions and manage tokens.
+
+<details>
```
tuliomir, line 68: thought: I really like this whole block, but I worry that it may become obsolete if we change the `src/api-docs.js` and forget to update it here as well.

For now, I'd prefer keeping this information duplicate as you just did, but it's a small flag we should keep in mind.

tuliomir (approved):
Technically this document is good to me. I'd only like to also ask for a glance from @BrunoCampana to consider this text in relation with our other documents.

docs/fireblocks.md
pedroferreira1: This should have been better detailed in the Acceptance Criteria. This is not supposed to be a complete and external guide. This is an internal guide that will be used by Campana to create the official tutorial, where I think we should have this explanation about Fireblocks

docs/fireblocks.md
```diff
@@ -0,0 +1,92 @@
+# Fireblocks integration
```
pedroferreira1, line 1: Since we are not installing the fireblocks package, I think we should say somewhere here the API version that we are currently using, if it exists.

pedroferreira1 (dismissed):

docs/fireblocks.md
r4mmer: Done

docs/fireblocks.md
r4mmer: Yes, but they don't have one for this.
RAW signing is a way to externally manage the wallet (which is what we are doing) so it depends on how the integration will be done, this is why I think they didn't document this.

tuliomir (dismissed):

pedroferreira1 (approved):

tuliomir (approved):

PR #450, opened 2024-06-12
feat: google cloud platform pubsub plugin integration

Ikari:
This PR have to changes:

- Google's pubsub integration: hathor_pubsub.js is a plugin that allows the wallet to send external notifications to a pubsub topic
- Fixing multisig param types: when deploying on some cloud environments (such as GCP Cloud Run), the **_multisig_seed_${x}_max_signatures_** and **_multisig_seed_${x}_num_signatures_** are threated as strings, not integers. When starting the multisig, the comparison made with this params is a exact on (  "===") so it fails. By parsing the params before we avoid that.

src/controllers/index.controller.js Ln 100-101:

```
  const mconfig = config.multisig[multisigKey];
    console.log('mconfig', mconfig);
    if (!(mconfig
           && (mconfig.total && mconfig.numSignatures && mconfig.pubkeys)
           && (mconfig.pubkeys.length === mconfig.total)
           && (mconfig.numSignatures <= mconfig.total))) {
      // Missing multisig items
      console.error(`Improperly configured multisig for seed ${multisigKey}.`);
      res.send({
        success: false,
        message: `Improperly configured multisig for seed ${multisigKey}.`
      });
      return;
```


config.js.docker
```diff
@@ -102,14 +102,21 @@ for (const x of seeds) {
...
     continue;
   }
+
+  // This parse is necessary cause some infrastructures treat all env variables as string. 
+  // As the comparision is for a equal value (===) that would make it fail. 
+  const multisig = {
+    total:          parseInt(multisigTotal),
+    numSignatures:  parseInt(multisigNumSigs),
+  };
+
```
luislhl, line 119: I believe using `getIntParam` would be enough and more concise. Also, it has some additional protections against a strange parseInt behavior.

```suggestion
  const multisig = {
    total:          getIntParam(`multisig_seed_${x}_max_signatures`),
    numSignatures:  getIntParam(`multisig_seed_${x}_num_signatures`),
  };
  if (!(multisigPubkeys || multisig.total || multisig.numSignatures)) {
    // No multisig parameters means this is not a multisig.
    continue;
  }
```

src/plugins/hathor_pubsub.js
```diff
@@ -0,0 +1,78 @@
+/**
```
r4mmer, line 1: You should add the config for this plugin on `src/plugins/child.js` to make it easier to enable.

r4mmer (changes requested):

src/plugins/hathor_pubsub.js
Ikari: Awesome. Just added the fix.

Ikari: Done

r4mmer (approved):

src/plugins/hathor_pubsub.js
```diff
@@ -0,0 +1,78 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/* istanbul ignore next */
+async function checkDeps() {
+  const requiredDeps = {
+    '@google-cloud/pubsub': '^4.0.1',
```
luislhl, line 11: I think having this dependency and not adding this package in the Docker image we send to Docker Hub is slightly odd.

Someone trying to use the plugin would be unable to, unless a custom image is built from ours that includes the needed package.

On the other hand, I'm not sure we want to always add the package just because of a plugin that a small percentage of the wallet-headless users will need.

I'm not sure the best way to deal with this. Maybe building a second Docker image that contains the package?

What do you think, @r4mmer @pedroferreira1 ?

pedroferreira1 (approved):

pedroferreira1:
@Ikari the PR is approved, it just needs a rebase so we can continue with the merge

luislhl (approved):

Ikari:
> @Ikari the PR is approved, it just needs a rebase so we can continue with the merge

Rebased!

pedroferreira1:
@Ikari we've fixed a bug in codecov, so we can make the CI succeed. Can you do one more rebase please, to update your branch, so we can check if the CI completes?

PR #461, opened 2024-07-08
feat: per wallet lock

r4mmer:
### Acceptance Criteria

- `SENT_TX` lock should be applied on a per-wallet basis

### Security Checklist

### DIFF
--- a/src/controllers/index.controller.js
+++ b/src/controllers/index.controller.js
@@ -295,6 +295,8 @@ function getConfigurationString(req, res) {
 }
 
 async function pushTxHex(req, res) {
+  // We don't need to use the send lock here since no shared resources are used.
+  // So this is a safe operation to run concurrently.
   const config = settings.getConfig();
 
   const validationResult = parametersValidation(req);
--- a/src/controllers/index.controller.js
+++ b/src/controllers/index.controller.js
@@ -8,13 +8,12 @@
 const { walletApi, tokensUtils, walletUtils, Network, helpersUtils, SendTransaction, constants: hathorLibConstants } = require('@hathor/wallet-lib');
 const { getApiDocs } = require('../api-docs');
 const { initializedWallets, startWallet } = require('../services/wallets.service');
-const { cantSendTxErrorMessage, API_ERROR_CODES } = require('../helpers/constants');
+const { API_ERROR_CODES } = require('../helpers/constants');
 const { parametersValidation } = require('../helpers/validations.helper');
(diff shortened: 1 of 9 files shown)


src/lock.js
```diff
@@ -68,4 +69,42 @@ class Lock {
   }
 }
 
-export const lock = new Lock();
+class GlobalLock {
+  constructor() {
+    this.wallets = {};
```
pedroferreira1, line 74: suggestion (blocking): We should add a comment explaining that this is to store a lock per wallet where the key is the wallet-id and the value is the `Lock` object.

src/lock.js
```diff
@@ -68,4 +69,42 @@ class Lock {
...
+    this.globalLock = new Lock();
+  }
+
+  start(walletId) {
```
pedroferreira1, line 78:
```suggestion
  _start(walletId) {
```

pedroferreira1: suggestion (non-blocking): this method shouldn't be used outside of the class.

src/lock.js
```diff
@@ -68,4 +69,42 @@ class Lock {
...
+
+  delete(walletId) {
+    if (this.wallets[walletId]) {
+      delete this.wallets[walletId];
+    }
+  }
+
+  get(walletId) {
+    if (!this.wallets[walletId]) {
+      return this.start(walletId);
```
pedroferreira1, line 93: suggestion (non-blocking): remove `start` method. We can have the line `this.wallets[walletId] = new Lock();` here and remove the `start` method. What do you think?

r4mmer: Done

src/lock.js
r4mmer: start method was removed

src/lock.js
r4mmer: Done

pedroferreira1 (approved):

pedroferreira1 (approved):

tuliomir (dismissed):

src/controllers/index.controller.js
```diff
@@ -304,12 +303,6 @@ async function pushTxHex(req, res) {
     return;
   }
 
-  const canStart = lock.lock(lockTypes.SEND_TX);
```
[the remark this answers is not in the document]
r4mmer: Done

tuliomir (approved):

pedroferreira1 (approved):

PR #462, opened 2024-07-22
Feat/streaming history sync

r4mmer:
### Acceptance Criteria

- Add configuration option for history sync mode
- Add history_sync_mode argument to start API.
  - If not present, fallback to history_sync_mode on config.

### Security Checklist

### DIFF
--- a/config.js.template
+++ b/config.js.template
@@ -106,4 +106,9 @@ module.exports = {
   // You should consider what components you are enabling in the /health endpoint when setting this.
   // Check the /health endpoint documentation for more info.
   considerHealthcheckWarnAsUnhealthy: true,
+
+  // Configure default history sync mode for the wallets
+  // Accepted values are: 'polling_http_api', 'xpub_stream_ws', 'manual_stream_ws'
+  // If no value is configured the default of 'polling_http_api' will be used
+  // history_sync_mode: 'polling_http_api|xpub_stream_ws|manual_stream_ws'
 };
--- a/src/services/wallets.service.js
+++ b/src/services/wallets.service.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-const { Connection, HathorWallet } = require('@hathor/wallet-lib');
+const { Connection, HathorWallet, HistorySyncMode } = require('@hathor/wallet-lib');
 const { removeAllWalletProposals } = require('./atomic-swap.service');
 const { notificationBus } = require('./notification.service');
 const { sanitizeLogInput } = require('../logger');
@@ -72,6 +72,7 @@ async function stopAllWallets() {
  * @param {Configuration} config Application configuration
  * @param {object} [options={}] Additional options
  * @param {string} [options.hsmKeyName] HSM key name
+ * @param {string} [options.historySyncMode] History sync mode
  * @returns {Promise<Object>} Returns the fullnode version data
  */
 async function startWallet(walletId, walletConfig, config, options = {}) {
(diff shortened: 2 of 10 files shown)


config.js.template
```diff
@@ -106,4 +106,9 @@ module.exports = {
   // You should consider what components you are enabling in the /health endpoint when setting this.
   // Check the /health endpoint documentation for more info.
   considerHealthcheckWarnAsUnhealthy: true,
-};
\ No newline at end of file
+
+  // Configure default history sync mode for the wallets
+  // Accepted values are: 'polling_http_api', 'xpub_stream_ws', 'manual_stream_ws'
+  // If no value is configured the default of 'polling_http_api' will be used
```
msbrogli, line 112: By default, should we try to use the xpub stream and fallback to the pooling if it's not available?

config.js.template
```diff
@@ -106,4 +106,9 @@ module.exports = {
...
+  // history_sync_mode: 'polling_http_api|xpub_stream_ws|manual_stream_ws'
```
msbrogli, line 113: Should we have a list of enabled sync modes with prioritization (e.g., an integer)?

src/services/wallets.service.js
```diff
@@ -94,6 +95,34 @@ async function startWallet(walletId, walletConfig, config, options = {}) {
...
+      case 'manual_stream_ws':
+        mode = HistorySyncMode.MANUAL_STREAM_WS;
+        break;
+      default:
+        break;
+    }
+
+    if (hydratedWalletConfig.multisig) {
+      // XXX: Multisig is not supported on streamming yet
+      mode = HistorySyncMode.POLLING_HTTP_API;
```
msbrogli, line 117: Maybe log what sync mode was used?

msbrogli (changes requested):

config.js.template
r4mmer: When introducing new features that work as an alternative to existing behavior I think keeping the default as the "current behavior" is better for people that constantly upgrade to the latest version since we keep their wallets working the same way since not using this config will not have any impact.

src/services/wallets.service.js
r4mmer: Done

config.js.template
msbrogli: I suggested this so a wallet configured to use `xpub_stream_ws` would fallback to `polling_http_api` if it connects to a full node where the selected mode is not available.

src/services/wallets.service.js
msbrogli: Thanks!

config.js.template
r4mmer: This [issue] (https://github.com/HathorNetwork/hathor-wallet-lib/issues/715) was create to track this

msbrogli: Ok. Thanks!

msbrogli (dismissed):

src/services/wallets.service.js
```diff
@@ -94,6 +95,35 @@ async function startWallet(walletId, walletConfig, config, options = {}) {
...
+        break;
+      case 'manual_stream_ws':
+        mode = HistorySyncMode.MANUAL_STREAM_WS;
+        break;
+      default:
+        break;
+    }
+
+    if (hydratedWalletConfig.multisig) {
+      // XXX: Multisig is not supported on streamming yet
```
glevco, line 116: Typo: `streaming`

src/services/wallets.service.js
```diff
@@ -94,6 +95,35 @@ async function startWallet(walletId, walletConfig, config, options = {}) {
...
+      mode = HistorySyncMode.POLLING_HTTP_API;
+    }
+    if (hydratedWalletConfig.scanPolicy?.policy && hydratedWalletConfig.scanPolicy?.policy !== 'gap-limit') {
+      // XXX: currently only gap-limit can use streamming modes
```
glevco, line 120: Typo: `streaming`

glevco (dismissed):

msbrogli (approved):

PR #473, opened 2024-08-21
chore: upgrade actions to a version that use node20 (LTS)

r4mmer:
### Acceptance Criteria
- upgrade actions to a version that use node20 (LTS)

### Security Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,22 +23,22 @@ jobs:
 
     steps:
       - name: Checkout
-        # https://github.com/actions/checkout/releases/tag/v3.5.0
-        uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
+        # https://github.com/actions/checkout/releases/tag/v4.1.7
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
 
       - name: Initialize CodeQL
-        # https://github.com/github/codeql-action/releases/tag/codeql-bundle-v2.15.0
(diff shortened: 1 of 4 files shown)


.github/workflows/codeql.yml
```diff
@@ -23,22 +23,22 @@ jobs:
...
       - name: Checkout
-        # https://github.com/actions/checkout/releases/tag/v3.5.0
-        uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
+        # https://github.com/actions/checkout/releases/tag/v4.1.7
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
 
       - name: Initialize CodeQL
-        # https://github.com/github/codeql-action/releases/tag/codeql-bundle-v2.15.0
-        uses: github/codeql-action/init@517782a2a0dd93543ea4f12f41006cf70ddca135
+        # https://github.com/github/codeql-action/tree/v3.26.3
```
tuliomir, line 30: question(non-blocking): Why use the `/tree` as a reference and not the `/releases`?

tuliomir (approved):

pedroferreira1 (approved):

.github/workflows/codeql.yml
r4mmer: Because the releases only make references to the bundles, not the action releases.
This is better viewed in the actions [readme version suggestions](https://github.com/github/codeql-action?tab=readme-ov-file#supported-versions-of-the-codeql-bundle-and-github-enterprise-server).

The tags mark the action version so I used the code tagged under tag `v3.26.3` which seems more appropriate.
I believe that the bundle version will always be the latest, so we were marking something that was not being used.

PR #477, opened 2024-08-23
feat: include the wallet-id responsible for the log

r4mmer:
### Acceptance Criteria
- Include the wallet-id on the logs where possible

### Security Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,6 +40,18 @@
   "ignorePatterns": [ "src/config.js", "src/config.*.js", "dist/*" ],
   "overrides": [
     {
+      "files": [ "src/**" ],
+      "rules": {
+        "no-console": "error"
+      }
+    },
+    {
+      "files": [ "src/plugins/**" ],
+      "rules": {
+        "no-console": "off"
+      }
+    },
+    {
       "files": [
         "**/*.js"
       ],
--- a/src/app.js
+++ b/src/app.js
@@ -14,6 +14,7 @@ import { ReadonlyErrorHandler } from './middlewares/xpub-error-handler.middlewar
 import { buildAppLogger } from './logger';
 import mainRouter from './routes/index.routes';
 import { initHathorLib } from './helpers/wallet.helper';
+import { loggerMiddleware } from './middlewares/logger.middleware';
 
 // Initializing Hathor Lib
 
@@ -27,6 +28,7 @@ const createApp = config => {
   app.use(express.json());
   app.use(express.urlencoded({ extended: true }));
   app.use(morgan(config.httpLogFormat || 'combined', { stream: logger.stream }));
+  app.use(loggerMiddleware);
 
   if (config.http_api_key) {
     app.use(apiKeyAuth(config.http_api_key));
@@ -35,8 +37,8 @@ const createApp = config => {
   app.use(mainRouter);
   app.use(ConfigErrorHandler);
   app.use(ReadonlyErrorHandler);
-  app.use((err, _req, res, next) => {
-    console.error(err.stack);
+  app.use((err, req, res, next) => {
+    req.logger.error(err.stack);
     res.status(err.statusCode || 500).json({ message: err.message, stack: err.stack });
   });
 
--- a/src/app.js
+++ b/src/app.js
@@ -11,7 +11,7 @@ import morgan from 'morgan';
 import apiKeyAuth from './middlewares/api-key-auth.middleware';
 import { ConfigErrorHandler } from './middlewares/config-error-handler.middleware';
 import { ReadonlyErrorHandler } from './middlewares/xpub-error-handler.middleware';
-import buildLogger from './logger';
+import { buildAppLogger } from './logger';
 import mainRouter from './routes/index.routes';
 import { initHathorLib } from './helpers/wallet.helper';
 
@@ -20,7 +20,7 @@ import { initHathorLib } from './helpers/wallet.helper';
 const createApp = config => {
   initHathorLib(config);
 
-  const logger = buildLogger(config);
+  const logger = buildAppLogger(config);
 
   // Initializing ExpressJS
   const app = express();
@@ -35,7 +35,7 @@ const createApp = config => {
   app.use(mainRouter);
   app.use(ConfigErrorHandler);
   app.use(ReadonlyErrorHandler);
-  app.use((err, req, res, next) => {
+  app.use((err, _req, res, next) => {
     console.error(err.stack);
     res.status(err.statusCode || 500).json({ message: err.message, stack: err.stack });
   });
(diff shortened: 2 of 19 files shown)


.eslintrc
```diff
@@ -39,6 +39,18 @@
...
+      "rules": {
+        "no-console": "error"
+      }
+    },
+    {
+      "files": [ "src/plugins/**" ],
+      "rules": {
+        "no-console": "off"
+      }
+    },
```
r4mmer, line 53: These 2 entries make the rule `no-console` an error on all source files except the plugins.
The console should be used on the scripts, tests and plugins directly but we are changing to use only from the logger when inside the wallet routes.

I prefer setting `no-console` as off by default so any new directory is not affected by this rule.

src/app.js
```diff
@@ -36,7 +38,7 @@ const createApp = config => {
   app.use(ConfigErrorHandler);
   app.use(ReadonlyErrorHandler);
   app.use((err, req, res, next) => {
-    console.error(err.stack);
+    req.logger.error(err.stack);
```
tuliomir, line 41: thought: The patch coverage for this PR is low and is not passing the CI. I believe one of the reasons is that it that it targets error handling codes from other PRs that were never covered: the scope of this PR overlaps with the technical debt of many others.

Probably it will be necessary to add tests increasing the coverage over the updated lines, increasing the original scope of this PR.

src/services/logger.service.js
```diff
@@ -0,0 +1,52 @@
...
+  } else {
+    // There is no wallet for this request
+    // will return the app logger instance.
+    return buildAppLogger();
+  }
+
+  const logger = walletLoggers.get(walletId);
+  if (!logger) {
+    // Uninitialized wallet, should use the app logger.
+    return buildAppLogger();
```
pedroferreira1, line 47: Can we have here the app logger with the walletId parameter? Maybe we could have a log with `non-initialized wallet {wallet-id}` or something like that in this case. What do you think?

src/logger.js
```diff
@@ -57,33 +63,61 @@ function buildLogger(config) {
...
-
-  return logger;
+  return appLogger;
 }
 
 // This fixes some logs where Github code scanning complains about a `Log Injection` possibility
-const sanitizeLogInput = input => String(input).replace(/\n|\r/g, '');
+function sanitizeLogInput(input) {
+  return String(input).replace(/\n|\r/g, '');
+}
```
pedroferreira1, line 110: Why did you change this?

src/logger.js
```diff
@@ -7,14 +7,20 @@
 
 import util from 'util';
 import winston from 'winston';
+import { getConfig } from './settings';
+/** @import { Configuration } from './settings' */
```
pedroferreira1, line 11: Should we remove this?

src/services/fireblocks.service.js
```diff
@@ -253,11 +253,13 @@ class FireblocksClient {
         && e.response.data
         && e.response.data.code === TX_ALREADY_EXISTS_ERROR_CODE
       )) {
+        // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
pedroferreira1: It would be good, if possible

src/helpers/tx.helper.js
```diff
@@ -226,10 +229,10 @@ async function getTx(wallet, id) {
       return response.tx;
     }
 
-    console.warn('Failed to get transaction from fullnode.', response.message);
+    logger.warn('Failed to get transaction from fullnode.', response.message);
```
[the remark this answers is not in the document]
pedroferreira1: The `AppLogger` as default is good

src/controllers/fireblocks/fireblocks.controller.js
```diff
@@ -63,7 +66,7 @@ async function startFireblocksWallet(req, res) {
   if (!(config.fireblocksUrl
         && config.fireblocksApiKey
         && (config.fireblocksApiSecret || config.fireblocksApiSecretFile))) {
-    console.error('Error starting wallet because fireblocks is not configured.');
+    logger('Error starting wallet because fireblocks is not configured.');
```
pedroferreira1, line 69:
```suggestion
    logger.error('Error starting wallet because fireblocks is not configured.');
```

src/helpers/tx.helper.js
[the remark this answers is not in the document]
r4mmer: >If we forget the optional parameter options.logger this line will crash, right?

But `options.logger` is not optional, the idea is that it always should be passed.

src/services/fireblocks.service.js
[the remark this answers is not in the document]
r4mmer: Since the wallet can only start 1 fireblocks wallet per instance I thought that the fireblocks specific logs would not create a doubt on which wallet it comes from.
But I guess it would be nicer to be more consistent with this logger, so I changed to include the logger instance

src/logger.js
r4mmer: We were having errors on the tests because `sanitizeLogInput` was not a function.
I believe that the issue was on how the exports were being organized, since we added more exports in the file it changed the behavior.

The method is the same, only that now it is a function.

src/logger.js
r4mmer: This is not a comment, it is a JSDoc import so we can use the `Configuration` defined as a `@typedef` in `src/settings.js`.

We can remove it, personally I just like having the argument types since it helps with development (LSP completion and checks) and it helps with documentation since it helps understand what is expected to be in the config object.

src/index.js
```diff
@@ -28,10 +28,12 @@ async function startHeadless() {
       process.argv.slice(2),
       { silent: true, env: process.env },
     );
+    // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
r4mmer: Done.
The application logger has not yet been initialized.
Maybe later we can change the logger initialization so all `console.log`s are from the logger.

src/app.js
r4mmer: The reason was that some tests were failing.
Once the issue was resolved the patch coverage came up to 87%

src/services/logger.service.js
r4mmer: Done

pedroferreira1 (approved):

src/helpers/tx.helper.js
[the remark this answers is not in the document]
tuliomir: It seems the only place this is used is inside the `decodeTx` controller handler. In this case, I think it's not worth having too much effort handling this problem.

But I think letting the javascript throw an object destructuring error is a sub-optimal approach to a method meant to serve as a helper.

tuliomir (approved):

PR #480, opened 2024-08-27
feat: custom logger for lib

r4mmer:
### Acceptance Criteria

- Configure the headless logger on the lib
- Bump lib with latest changes on version v1.12.0

### Security Checklist

### DIFF
--- a/src/services/logger.service.js
+++ b/src/services/logger.service.js
@@ -6,6 +6,7 @@
  */
 
 import { buildAppLogger, buildServiceLogger } from '../logger';
+import { HathorWallet } from '@hathor/wallet-lib';
 
 /**
  * Wallet loggers mapped by walletId
@@ -37,7 +38,7 @@ function buildLibLogger(logger) {
 /**
  * Initialize a wallet logger
  * @param {string} walletId
- * @returns {import('winston').Logger}
+ * @returns {[import('winston').Logger, ILogger]}
  */
 function initializeWalletLogger(walletId) {
   const logger = buildServiceLogger(`wallet(${walletId})`);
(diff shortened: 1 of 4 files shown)


src/services/logger.service.js
```diff
@@ -48,4 +72,41 @@ function getLogger(req) {
...
+        logger.debug(`state_update[syncing]: time to connect ${(times.syncing - times.connecting) / 1000} seconds`);
+        break;
+      case HathorWallet.PROCESSING:
+        times.processing = Date.now();
+        logger.debug(`state_update[processing]: time to sync ${(times.processing - times.syncing) / 1000} seconds`);
+        break;
+      case HathorWallet.READY:
+        times.ready = Date.now();
+        logger.debug(`state_update[ready]: time to process history ${(times.ready - times.processing) / 1000} seconds`);
+        break;
```
tuliomir, line 102: question: Should we remove the listener on wallet ready? It doesn't seem to be used anymore.

src/services/logger.service.js
```diff
@@ -48,4 +72,41 @@ function getLogger(req) {
...
+      default:
+        break;
+    }
+  });
+  wallet.conn.on('wallet-load-partial-update', data => {
+    logger.debug(`Found ${data.addressesFound} addresses and ${data.historyLength} transactions`);
```
tuliomir, line 108: question: Do we have any moment when this listener is no longer needed? If so, can we remove it at this time?

src/services/logger.service.js
r4mmer: It will be used if the connection is restarted due to some instability.
The logs will then reassign the values and log the time to become ready again.

Leaving this on is a feature that will help understand better how the wallets react when starting for the first time and during a restart.

src/services/logger.service.js
r4mmer: This event is emitted when the wallet is syncing, since we may sync again if the connection is lost we should leave this on.

tuliomir (dismissed):

luislhl (approved):

tuliomir (approved):

luislhl (approved):

PR #484, opened 2024-09-16
feat: mark inputs as used

r4mmer:
### Acceptance Criteria
- Create a new API to mark utxos as selected for input or unmark them.
- Add an option on the p2sh proposal api to mark utxos as selected for input.

### Security Checklist

### DIFF
--- a/__tests__/mark_utxos_selected_as_input.test.js
+++ b/__tests__/mark_utxos_selected_as_input.test.js
@@ -22,7 +22,7 @@ function createCustomTxHex() {
 describe('mark utxos selected_as_input api', () => {
   let selectSpy;
   const txHex = createCustomTxHex();
-  
+
   beforeAll(async () => {
     selectSpy = jest.spyOn(Storage.prototype, 'utxoSelectAsInput');
     await TestUtils.startWallet({ walletId });
@@ -107,8 +107,8 @@ describe('mark utxos selected_as_input api', () => {
 
     expect(selectSpy).toHaveBeenCalledTimes(3);
     expect(selectSpy).toHaveBeenCalledWith(
-      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74'}, 
-      true, 
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      true,
       undefined,
     );
   });
@@ -123,8 +123,8 @@ describe('mark utxos selected_as_input api', () => {
 
     expect(selectSpy).toHaveBeenCalledTimes(3);
     expect(selectSpy).toHaveBeenCalledWith(
-      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74'}, 
-      true, 
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      true,
       123,
     );
   });
@@ -139,8 +139,8 @@ describe('mark utxos selected_as_input api', () => {
 
     expect(selectSpy).toHaveBeenCalledTimes(3);
     expect(selectSpy).toHaveBeenCalledWith(
-      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74'}, 
-      false, 
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      false,
       undefined,
     );
   });
@@ -155,8 +155,8 @@ describe('mark utxos selected_as_input api', () => {
 
     expect(selectSpy).toHaveBeenCalledTimes(3);
     expect(selectSpy).toHaveBeenCalledWith(
-      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74'}, 
-      true, 
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      true,
       undefined,
     );
   });
@@ -171,8 +171,8 @@ describe('mark utxos selected_as_input api', () => {
 
     expect(selectSpy).toHaveBeenCalledTimes(3);
     expect(selectSpy).toHaveBeenCalledWith(
-      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74'}, 
-      false, 
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      false,
       456,
     );
   });
--- a/__tests__/mark_utxos_selected_as_input.test.js
+++ b/__tests__/mark_utxos_selected_as_input.test.js
@@ -1,0 +1,179 @@
+import { Storage, Transaction, Input, Output } from '@hathor/wallet-lib';
+import TestUtils from './test-utils';
+
+const walletId = 'stub_mark_inputs_as_used';
+
+function createCustomTxHex() {
+  const txId0 = '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74';
+  const txId1 = 'fb2fbe0385bc0bc8e9a255a8d530f7b3bdcebcd5ccdae5e154e6c3d57cbcd143';
+  const txId2 = '11835fae291c60fc58314c61d27dc644b9e029c363bbe458039b2b0186144275';
+  const tx = new Transaction(
+    [new Input(txId0, 0), new Input(txId1, 1), new Input(txId2, 2)],
+    [new Output(100, Buffer.from('0463616665ac', 'hex'))],
+    {
+      timestamp: 123,
+      parents: ['f6c83e3641a08ec21aebc01296ff12f5a46780f0fbadb1c8101309123b95d2c6'],
+    },
+  );
+
+  return tx.toHex();
+}
+
+describe('mark utxos selected_as_input api', () => {
+  let selectSpy;
+  const txHex = createCustomTxHex();
+  
+  beforeAll(async () => {
+    selectSpy = jest.spyOn(Storage.prototype, 'utxoSelectAsInput');
+    await TestUtils.startWallet({ walletId });
+  });
+
+  beforeEach(() => {
+    selectSpy.mockReset();
+    selectSpy.mockImplementation(jest.fn(async () => {}));
+  });
+
+  afterAll(async () => {
+    selectSpy.mockRestore();
+    await TestUtils.stopWallet({ walletId });
+  });
+
+  it('should fail if txHex is not a hex string', async () => {
+    let response = await TestUtils.request
+      .put('/wallet/utxos-selected-as-input')
+      .send({ txHex: 123 })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
+
+    response = await TestUtils.request
+      .put('/wallet/utxos-selected-as-input')
+      .send({ txHex: '0123g' })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
+    expect(selectSpy).toHaveBeenCalledTimes(0);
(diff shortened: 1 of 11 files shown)


__tests__/mark_utxos_selected_as_input.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+  return tx.toHex();
+}
+
+describe('mark utxos selected_as_input api', () => {
+  let selectSpy;
+  const txHex = createCustomTxHex();
+
+  beforeAll(async () => {
+    selectSpy = jest.spyOn(Storage.prototype, 'utxoSelectAsInput');
```
alexruzenhack, line 27: [question] Is it necessary to target the `prototype` or are you being extra explicit?

__tests__/mark_utxos_selected_as_input.test.js
```diff
@@ -0,0 +1,179 @@
...
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
+
+    response = await TestUtils.request
+      .put('/wallet/utxos-selected-as-input')
+      .send({ txHex, mark_as_used: 'abc' })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
```
alexruzenhack, line 78: [question] A string value for `mark_as_used` was already tested above, why test it again?

__tests__/mark_utxos_selected_as_input.test.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+
+  it('should fail if ttl is not a number', async () => {
+    const response = await TestUtils.request
+      .put('/wallet/utxos-selected-as-input')
+      .send({ txHex, ttl: '123a' })
+      .set({ 'x-wallet-id': walletId });
+    expect(response.status).toBe(400);
+    expect(response.body.success).toBe(false);
```
alexruzenhack, line 87: [question] Should we test a number as string?

__tests__/mark_utxos_selected_as_input.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+    expect(selectSpy).toHaveBeenCalledTimes(3);
+    expect(selectSpy).toHaveBeenCalledWith(
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      false,
+      undefined,
+    );
+  });
+
+  it('should mark the inputs as selected on the storage with mark true', async () => {
```
alexruzenhack, line 148: [question] This is the default behavior, right?

__tests__/mark_utxos_selected_as_input.test.js
```diff
@@ -0,0 +1,179 @@
...
+      { index: 0, txId: '5db0a8c77f818c51cb107532fc1a36785adfa700d81d973fd1f23438b2f3dd74' },
+      true,
+      undefined,
+    );
+  });
+
+  it('should mark the inputs as selected on the storage with all options', async () => {
+    const response = await TestUtils.request
+      .put('/wallet/utxos-selected-as-input')
+      .send({ txHex, mark_as_used: false, ttl: 456 })
```
alexruzenhack, line 167: [question] What does it means?

__tests__/p2sh/tx-proposal-create.test.js
```diff
@@ -151,4 +151,31 @@ describe('create tx-proposal api', () => {
...
+        .set({ 'x-wallet-id': walletId });
+      expect(response.status).toBe(200);
+      console.log(JSON.stringify(response.body));
+      expect(response.body.txHex).toBeDefined();
+      expect(response.body.success).toBe(true);
+      const tx = hathorLib.helpersUtils.createTxFromHex(response.body.txHex, new hathorLib.Network('testnet'));
+      expect(tx.outputs.map(o => o.decodedScript.address.base58))
+        .toEqual(expect.arrayContaining(['WPynsVhyU6nP7RSZAkqfijEutC88KgAyFc', 'wcUZ6J7t2B1s8bqRYiyuZAftcdCGRSiiau']));
+
+      expect(selectSpy).toHaveBeenCalledTimes(1);
```
alexruzenhack, line 176: [question] Should we extend the expect tests to check also the call arguments?

src/routes/wallet/wallet.routes.js
```diff
@@ -465,4 +466,45 @@ walletRouter.post(
...
+  '/utxos-selected-as-input',
+  checkSchema({
+    txHex: {
+      in: ['body'],
+      errorMessage: 'Invalid txHex',
+      isString: true,
+      custom: {
+        options: (value, { req, location, path }) => {
+          // Test if txHex is actually hex
+          if (!(/^[0-9a-fA-F]+$/.test(value))) return false;
```
alexruzenhack, line 483: [question] We doesn't need to challenge the length?

src/controllers/wallet/wallet.controller.js
```diff
@@ -763,6 +763,42 @@ async function stop(req, res) {
...
+  /** @type {{ wallet: import('@hathor/wallet-lib').HathorWallet }} */
+  const { wallet } = req;
+  const { txHex, ttl } = req.body;
+  const markAsUsed = req.body.mark_as_used ?? true;
+
+  try {
+    const tx = helpersUtils.createTxFromHex(txHex, wallet.getNetworkObject());
+    await Promise.all(
+      tx.inputs.map(
+        input => (wallet.storage.utxoSelectAsInput(
```
alexruzenhack, line 788: [note] I believe there is an issue in the `utxoSelectAsInput`:

```js
  /**
   * Mark an utxo as selected as input
   *
   * @param {IUtxoId} utxo The Data to identify the utxo
   * @param {boolean} markAs Mark the utxo as this value
   * @param {number|undefined} ttl Unmark the utxo after this amount os ms passed
   *
   * @async
   * @returns {Promise<void>}
   */
  async utxoSelectAsInput(utxo: IUtxoId, markAs: boolean, ttl?: number): Promise<void> {
    const tx = await this.getTx(utxo.txId);
    if (!tx || !tx.outputs[utxo.index]) {
      return;
    }

    if (markAs && tx.outputs[utxo.index].spent_by !== null) {
      // Already spent, no need to mark as selected_as_input
      return;
    }

    const utxoId = `${utxo.txId}:${utxo.index}`;
    if (markAs) {
      this.utxosSelectedAsInput.set(utxoId, markAs);
      // if a ttl is given, we should reverse
      if (ttl) {
        setTimeout(() => {
          if (!markAs) {
            if (this.utxosSelectedAsInput.has(utxoId)) {
              this.utxosSelectedAsInput.delete(utxoId);
            }
          }
        }, ttl);
      }
    } else {
      this.utxosSelectedAsInput.delete(utxoId);
    }
  }
```

If `markAs` is `false` or `undefined` (I know we are validating to never let it be `undefined` from headless), we will delete the utxo as selected input, but we will not restore its previous state if we use it with `ttl` because there is no `setTimeout` in the `else` branch. Is it an issue? Do we want this behavior of ignoring `ttl` in case `markAs` is `false`?

__tests__/mark_utxos_selected_as_input.test.js
r4mmer: It is necessary because in jest, `jest.spyOn(Storage, 'utxoSelectAsInput')` would only work if `utxoSelectAsInput` was a static method.
Since it is a normal method this mock would not affect the instances of `Storage`.
To affect the instances of `Storage` we need to mock the `Storage.prototype.utxoSelectAsInput`.

__tests__/mark_utxos_selected_as_input.test.js
r4mmer: Yes, but this test is to check that the behavior does not change when we send `mark_as_used` as `true`

__tests__/p2sh/tx-proposal-create.test.js
r4mmer: We can but in this specific test we dont know the input being selected.
It should select one from the history fixture and if we just copy the details and the fixture changes in the future, the test would fail without explanation.

src/controllers/wallet/wallet.controller.js
r4mmer: This is not a bug, we intentionally do not revert an unmark as "selected_as_input".

But while checking the method I found a bug on the method, created another [PR](https://github.com/HathorNetwork/hathor-wallet-lib/pull/773) to fix this

alexruzenhack: Oh, ok!

__tests__/mark_utxos_selected_as_input.test.js
alexruzenhack: Interesting! Thank you for the knowledge.

alexruzenhack (approved):

src/routes/wallet/wallet.routes.js
r4mmer: We don't know the length of the txHex.
We know that the length should be higher than 10 chars (5 bytes) and the ceiling is a few kilobytes (do not know the exact number) but it is so variable that there is no need to check the length.

alexruzenhack: Got it!

src/controllers/wallet/p2sh/tx-proposal.controller.js
```diff
@@ -48,6 +50,10 @@ async function buildTxProposal(req, res) {
       changeAddress,
     });
     const txData = await sendTransaction.prepareTxData();
+
+    if (markAsUsed) {
+      await sendTransaction.updateOutputSelected(true);
+    }
```
pedroferreira1, line 56: Does it need another PR from the lib? I might be wrong but seeing the flow from the lib, the `updateOutputSelected` checks `if (this.transaction === null) {` and, in this case, it should be `null`.

```
const sendTransaction = new SendTransaction({
      storage: req.wallet.storage,
      outputs,
      inputs,
      changeAddress,
    });
    const txData = await sendTransaction.prepareTxData();
```

You instantiate the object with `outputs` and `inputs` and the method `prepareTxData` doesn't set `this.transaction`

src/routes/wallet/p2sh/tx-proposal.routes.js
```diff
@@ -81,6 +81,12 @@ txProposalRouter.post(
       isString: true,
       optional: true,
     },
+    mark_inputs_as_used: {
+      in: ['body'],
+      errorMessage: 'Invalid mark_as_used argument',
+      isBoolean: true,
```
pedroferreira1, line 87: Should we have `toBoolean: true` here as well?

src/routes/wallet/wallet.routes.js
```diff
@@ -465,4 +466,45 @@ walletRouter.post(
  */
 walletRouter.post('/stop', stop);
 
+/**
+ * PUT request to mark inputs from a txHex as used.
+ * For the docs, see api-docs.js
+ */
+walletRouter.put(
+  '/utxos-selected-as-input',
```
pedroferreira1, line 474: In the `atomic-swap` routes we have a simular router, `/unlock`, which does the same for an atomic swap `PartialTx`. I was thinking about this and if it would be better to unify both APIs. 

What do you think?

src/controllers/wallet/wallet.controller.js
pedroferreira1, line 788: Using storage methods directly seems not good to me, we have a method in the wallet class `markUtxoSelected` that we could use for that. The only downside is that we don't have the `ttl` parameter there.

What do you think would be better? 

1. Remove the `ttl` parameter here.
2. Add the `ttl` parameter in the lib.
3. Keep using the `storage` method directly?

pedroferreira1 (commented):
It's missing the `api-docs.js` change

src/routes/wallet/p2sh/tx-proposal.routes.js
r4mmer: Done

src/controllers/wallet/wallet.controller.js
pedroferreira1: For me it's good

src/routes/wallet/wallet.routes.js
r4mmer: I think that unifying these APIs would be the best choice, but I think we should do this on another PR.
Because deprecating the unlock API will require changing the docs and the website tutorial, maybe we should organize this on another issue.

What do you think?

alexruzenhack (approved):

src/routes/wallet/wallet.routes.js
pedroferreira1: I agree. Can you create the issue for this, please?

pedroferreira1 (approved):

PR #498, opened 2024-10-09
feat: early send tx release for wallet routes

r4mmer:
### Acceptance Criteria
- Early send-tx release 

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "@babel/node": "7.23.9",
     "@babel/plugin-proposal-class-properties": "7.18.6",
     "@babel/preset-env": "7.24.0",
+    "@types/express": "^5.0.0",
     "axios-mock-adapter": "1.22.0",
     "babel-jest": "29.7.0",
     "eslint": "8.57.0",
--- a/src/controllers/wallet/wallet.controller.js
+++ b/src/controllers/wallet/wallet.controller.js
@@ -11,9 +11,23 @@ const { matchedData } = require('express-validator');
 const { parametersValidation } = require('../../helpers/validations.helper');
 const { lock, lockTypes } = require('../../lock');
 const { cantSendTxErrorMessage, friendlyWalletState } = require('../../helpers/constants');
-const { mapTxReturn, prepareTxFunds, getTx, markUtxosSelectedAsInput } = require('../../helpers/tx.helper');
+const { mapTxReturn, prepareTxFunds, getTx, markUtxosSelectedAsInput, runSendTransaction } = require('../../helpers/tx.helper');
 const { stopWallet } = require('../../services/wallets.service');
 
+/**
+ * @typedef {import('@hathor/wallet-lib').SendTransaction} SendTransaction
+ * @typedef {import('@hathor/wallet-lib').HathorWallet} HathorWallet
+ * @typedef {import('winston').Logger} Logger
+ * @typedef {{ walletId: string, wallet: HathorWallet, logger: Logger }} HathorRequestExtras
+ * @typedef {import('express').Request} ExpressRequest
+ * @typedef {ExpressRequest & HathorRequestExtras} Request
+ * @typedef {import('express').Response} Response
+ */
+
+/**
+ * @param {Request} req
+ * @param {Response} res
+ */
 async function getStatus(req, res) {
   /**
    * @type {HathorWallet} wallet - Wallet object
@@ -239,6 +253,10 @@ async function getTxConfirmationBlocks(req, res) {
   res.send({ success: true, confirmationNumber });
 }
 
+/**
+ * @param {Request} req
+ * @param {Response} res
+ */
 async function simpleSendTx(req, res) {
   const validationResult = parametersValidation(req);
   if (!validationResult.success) {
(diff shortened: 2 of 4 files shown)


__tests__/__fixtures__/http-fixtures.js
```diff
@@ -706,6 +706,19 @@ export default {
...
+            script: 'dqkUsmE9zshkgB58piBD7ETUV0e/NgmIrA==',
+            decoded: {
+              type: 'P2PKH',
+              address: 'WewDeXWyvHP7jJTs7tjLoQfoB72LLxJQqN',
+              timelock: null,
+            },
+            token: '0000073b972162f70061f61cf0082b7a47263cc1659a05976aca5cd01b3351ee',
+            spent_by: null,
+            selected_as_input: false,
+          },
```
r4mmer, line 721: One of the fixes in the wallet-lib was to only select available authorities.
The only authority for this token is timelocked so some mint tests were failing.

src/controllers/wallet/wallet.controller.js
```diff
@@ -11,9 +11,23 @@ const { matchedData } = require('express-validator');
...
 
+/**
+ * @typedef {import('@hathor/wallet-lib').SendTransaction} SendTransaction
+ * @typedef {import('@hathor/wallet-lib').HathorWallet} HathorWallet
+ * @typedef {import('winston').Logger} Logger
+ * @typedef {{ walletId: string, wallet: HathorWallet, logger: Logger }} HathorRequestExtras
+ * @typedef {import('express').Request} ExpressRequest
+ * @typedef {ExpressRequest & HathorRequestExtras} Request
+ * @typedef {import('express').Response} Response
+ */
```
r4mmer, line 25: These are just some JSDoc definitions for documentation.

src/controllers/wallet/wallet.controller.js
```diff
@@ -239,21 +253,30 @@ async function getTxConfirmationBlocks(req, res) {
...
-  /**
-   * @type {{wallet: HathorWallet, logger: import('winston').Logger}}
-   */
+  let lockReleased = false;
+  const unlock = () => {
+    if (!lockReleased) {
+      walletLock.unlock(lockTypes.SEND_TX);
+    }
+    lockReleased = true;
+  };
```
r4mmer, line 279: This `unlock` function was made to avoid calling `walletLock.unlock` twice since we call unlock after the `prepare-tx` phase and also call in the finally.

The unlock in the `finally` is required because if any errors happens before the `runSendTransaction` is called we need to unlock the send-tx lock.

Since both `unlock`s are required and calling it for a second time may cause issues we need this method that can only be called once.

__tests__/mint-tokens.test.js
```diff
@@ -1,13 +1,14 @@
...
 
 describe('mint-tokens api', () => {
-  beforeAll(async () => {
+  beforeEach(async () => {
     await TestUtils.startWallet({ walletId, preCalculatedAddresses: TestUtils.addresses });
   });
 
-  afterAll(async () => {
+  afterEach(async () => {
     await TestUtils.stopWallet({ walletId });
```
r4mmer, line 12: These have changed because the only mint authority was being marked as "selected_as_input" and another is not being created (due to how the http mock works) so we have to restart the wallet on every test

src/controllers/wallet/atomic-swap/tx-proposal.controller.js
```diff
@@ -323,12 +323,12 @@ async function signAndPush(req, res) {
...
 
     const sendTransaction = new SendTransaction({ transaction, storage: req.wallet.storage });
-    const response = await sendTransaction.runFromMining();
+    const response = await runSendTransaction(sendTransaction, unlock);
     res.send({ success: true, ...mapTxReturn(response) });
   } catch (err) {
     res.send({ success: false, error: err.message });
   } finally {
-    lock.get(req.walletId).unlock(lockTypes.SEND_TX);
+    unlock();
```
pedroferreira1, line 331: The `unlock` method is always called in case there are no errors inside the `runSendTransaction` method. Should we move this call inside the `catch` block?

If yes, we should apply to all controllers.

src/helpers/lock.helper.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param {string} walletId
+ * @return {[CallableFunction, string|null]} The unlock function and the error message.
+ *
+ */
+function lockSendTx(walletId) {
+  const walletLock = lock.get(walletId);
+  const canStart = walletLock.lock(lockTypes.SEND_TX);
+  if (!canStart) {
+    return [() => {}, cantSendTxErrorMessage];
```
andreabadesso, line 22: Maybe you should return `null` instead of a noop so that if a dev forgets to check if the message is null he gets an error when trying to unlock

package.json
```diff
@@ -58,6 +58,7 @@
     "@babel/node": "7.23.9",
     "@babel/plugin-proposal-class-properties": "7.18.6",
     "@babel/preset-env": "7.24.0",
+    "@types/express": "^5.0.0",
```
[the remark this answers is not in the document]
r4mmer: Done, also updated the package-lock with this

src/helpers/tx.helper.js
```diff
@@ -252,10 +256,33 @@ async function markUtxosSelectedAsInput(wallet, utxos, markAs, ttl) {
...
+ * @returns {Promise<Transaction>}
+ */
+async function runSendTransaction(sendTx, unlock) {
+  try {
+    if (!sendTx.transaction) {
+      await sendTx.prepareTx();
+    }
+    await sendTx.updateOutputSelected(true);
+  } finally {
+    // If an error happens of things go as planned we release the lock
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/melt-tokens.test.js
```diff
@@ -22,6 +23,7 @@ describe('melt-tokens api', () => {
         amount: 1,
       })
       .set({ 'x-wallet-id': walletId });
+    console.log(JSON.stringify(response.body, null, 2));
```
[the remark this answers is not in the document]
r4mmer: Done

src/helpers/lock.helper.js
r4mmer: Done

pedroferreira1 (dismissed):

andreabadesso (dismissed):

andreabadesso (approved):

pedroferreira1 (approved):

PR #499, opened 2024-10-10
feat: add missing parameters in nano contract API

pedroferreira1:
### Acceptance Criteria
- Add `before` query parameter to the nano contract history API.
- Add `block_hash` and `block_height` query parameters to the nano contract state API.

### Security Checklist

### DIFF
--- a/src/controllers/wallet/nano-contracts.controller.js
+++ b/src/controllers/wallet/nano-contracts.controller.js
@@ -21,10 +21,24 @@ async function getState(req, res) {
     return;
   }
 
-  const { id, fields, balances, calls } = req.query;
+  const {
+    id,
+    fields,
+    balances,
+    calls,
+    block_hash: blockHash,
+    block_height: blockHeight
+  } = req.query;
 
   try {
-    const state = await ncApi.getNanoContractState(id, fields, balances, calls);
+    const state = await ncApi.getNanoContractState(
+      id,
+      fields,
+      balances,
+      calls,
+      blockHash,
+      blockHeight
+    );
 
     res.send({
       success: true,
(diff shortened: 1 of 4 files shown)


r4mmer (approved):

src/controllers/wallet/nano-contracts.controller.js
```diff
@@ -45,10 +59,10 @@ async function getHistory(req, res) {
...
 
-  const { id, count, after } = req.query;
+  const { id, count, after, before } = req.query;
 
   try {
-    const data = await ncApi.getNanoContractHistory(id, count, after);
+    const data = await ncApi.getNanoContractHistory(id, count, after, before);
 
     res.send({
       success: true,
```
luislhl, line 68: The changes in the PR seem good. But I also took the opportunity to check the full interaction from wallet-headless to wallet-lib to hathor-core.

It seems we could be losing information about some specific errors coming from hathor-core, which could make debugging harder when running wallet-headless.

This is a historic problem in wallet-headless that I think we should avoid in these new endpoints.

I didn't make an extensive review of all cases, but one of them would be when hathor-core returns this from the nano-contract state endpoint:

```
        except NCContractCreationVoided:
            request.setResponseCode(204)
            error_response = ErrorResponse(success=False, error=f'Nano contract failed execution: {params.id}')
            return error_response.json_dumpb()
```

What the lib will apparently do with this is:

```
      throw new NanoRequestError('Error getting nano contract state.', null, response);
```

Then, we will just return `err.message` to the user, which is `Error getting nano contract state.`. But the original error from hathor-core, which is `Nano contract failed execution`, will get missing.

Does this make sense?

luislhl (approved):

src/controllers/wallet/nano-contracts.controller.js
pedroferreira1: You are correct. This `204` error was added after the first integration of nano contract with the full node, that's why it was not handled properly.

Apart from that, the nano execution in the mempool is still under discussion, then some error codes might change. I created a task to improve this in the lib (the headless doesn't need any changes).

https://github.com/HathorNetwork/hathor-wallet-lib/issues/795

r4mmer (approved):

luislhl (approved):

PR #509, opened 2024-11-20
Feat/bigint

glevco:
### Acceptance Criteria
- Include here all things that this PR should solve

### Security Checklist


package.json
```diff
@@ -4,13 +4,13 @@
...
   "engines": {
-    "node": ">=20.0.0",
+    "node": ">=22.0.0",
     "npm": ">=10.0.0"
   },
   "dependencies": {
     "@dinamonetworks/hsm-dinamo": "4.9.1",
     "@hathor/healthcheck-lib": "0.1.0",
-    "@hathor/wallet-lib": "1.14.0",
+    "@hathor/wallet-lib": "git://github.com/HathorNetwork/hathor-wallet-lib#glevco/fake-release",
```
glevco, line 13: TODO: Change this to the actual released version, when we do it.

src/helpers/tx.helper.js
```diff
@@ -69,7 +69,12 @@ async function getUtxosToFillTx(wallet, sumOutputs, options) {
 
   const { utxos } = utxosDetails;
   // Sort utxos with larger amounts first
-  utxos.sort((a, b) => b.amount - a.amount);
+  utxos.sort((a, b) => {
+    if (a.amount < b.amount) {
+      return 1;
+    }
+    return a.amount > b.amount ? -1 : 0;
+  });
```
glevco, line 77: I had to change this because the output of `sort` expects a `Number`, so the previous code would fail. I also couldn't use nested ternary because the linter prevents it. But the behavior should be the same.

pedroferreira1 (approved):

src/helpers/tx.helper.js
r4mmer: Maybe we should add a test to this.
It would also increase the coverage of the PR

src/routes/wallet/p2sh/tx-proposal.routes.js
```diff
@@ -42,7 +43,9 @@ txProposalRouter.post(
           min: 1,
         },
       },
-      toInt: true,
+      customSanitizer: {
+        options: bigIntSanitizer,
+      },
```
r4mmer, line 48: question(non-blocking): We are going to remove the express-validation dependency correct?

Just checking since we can use zod for validation we could just remove these schemas, but I think this will be in another PR correct?

src/schemas.js
```diff
@@ -533,3 +548,13 @@ export const nanoContractData = {
...
+
+export function bigIntSanitizer(value) {
+  try {
+    return value === undefined || value === null
+      ? value
+      : parseSchema(value, bigIntCoercibleSchema);
+  } catch {
+    return undefined;
+  }
+}
```
r4mmer, line 560: suggestion: We should add some simple tests for this.

src/routes/wallet/p2sh/tx-proposal.routes.js
glevco: Correct, this is tracked in this issue: https://github.com/HathorNetwork/hathor-wallet-headless/issues/510

src/controllers/wallet/wallet.controller.js
```diff
@@ -510,7 +511,7 @@ async function sendTx(req, res) {
     const ret = { success: false, error: err.message };
     if (debug) {
       logger.debug('/send-tx failed', {
-        body: JSON.stringify(req.body),
+        body: JSONBigInt.stringify(req.body),
```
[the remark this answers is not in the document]
glevco: This is a good question, I don't think it would create a problem per-se, but I haven't tested it.

However, consider the JSON section in mdn's BigInt [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json):

> `JSON.stringify()` specifically leaves a backdoor for BigInt values: it would try to call the BigInt's `toJSON() `method. (It doesn't do so for any other primitive values.) Therefore, you can implement your own `toJSON() `method (which is one of the few cases where patching built-in objects is not explicitly discouraged).

Notice how it states that patching built-in objects is explicitly discouraged. You can find multiple other sources advocating against it. So, I wouldn't be too keen on patching `JSON.parse` or `JSON.stringify` directly, but we _could_ patch `BigInt.prototype.toJSON` as suggested in the docs. The second alternative, also explained there, is to use a `replacer`.

I ended up going with the latter because for the inverse case on `JSON.parse`, using a `reviver` is the only option. So to keep operations symmetric, I decided to use the same solution in both cases. Also, it makes things explicit rather than implicit.

> The reason I ask this is because in the future a dev may just use JSON without understanding the implications and introduce a bug that may only be caught if we use specific values during testing

I agree this could be problematic, but keep in mind that in the vast majority of cases the values will already be `BigInts`, even if they're small. I mean, whenever you're dealing with amounts, even if it's `1n`, it'll be a `BigInt`. The only case this would be a problem is when we have some object with a value that could be _either_ a `BigInt` or a `Number`, but I don't think we should have this case. Unfortunately this whole bigint deal is a bit of a pain in JS, and I think we should make an effort to keep this in mind whenever we're programming or reviewing. Later, using typing (either with TypeScript or just linters) will help prevent this kind of problem.

src/helpers/tx.helper.js
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/9c1caefcf3b04b7b9ba371d35d77643a5d5a0338

src/app.js
```diff
@@ -25,6 +27,8 @@ const createApp = config => {
 
   // Initializing ExpressJS
   const app = express();
+  app.set('json replacer', bigIntUtils.JSONBigInt.bigIntReplacer);
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/9c1caefcf3b04b7b9ba371d35d77643a5d5a0338

src/schemas.js
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/9c1caefcf3b04b7b9ba371d35d77643a5d5a0338

r4mmer (dismissed):

src/controllers/wallet/wallet.controller.js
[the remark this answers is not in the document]
r4mmer: The reviver we have only applies to API inputs, correct?
If we were to serialize/parse from anywhere else we would need to add this `toJSON` patch and use a reviver on the parse calls manually.

There can be a way to prevent `JSON.parse` without `reviver` as a custom eslint rule.
May be too much but checking `CallExpression:exit` where `JSON.parse(...)` only has 1 argument should be enough.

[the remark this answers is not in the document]
glevco: I created an issue for this: https://github.com/HathorNetwork/hathor-wallet-headless/issues/511

pedroferreira1 (dismissed):

package.json
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/02de65e83931e3580415cd9ca0b0228f0a1453a9

pedroferreira1 (dismissed):

glevco:
I had to update `axios` to the same version as the wallet-lib after the update, in commit https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/4bf1f9bc2bc4e7245394c4a170ba3b1f000c4d01.

Then, in commit https://github.com/HathorNetwork/hathor-wallet-headless/pull/509/commits/331d7ad41025f87c5f044b09782e4e17474389ac, I fixed an issue that I caught in manual testing. I had configured the JSON replacer for Express, but not the reviver. Then, I also had to change validation to work with `BigInt`.

r4mmer (approved):

pedroferreira1 (approved):

PR #517, opened 2024-12-23
fix: exit after scripts

r4mmer:
### Acceptance Criteria
- Call `process.exit` after a complete execution of any script

### Security Checklist


BrunoCampana (dismissed):

.nvmrc
```diff
@@ -1 +1 @@
-v22
+v23
```
tuliomir, line 1: issue: I believe we should not use unstable versions of node for the headless. Is there any benefit in this case?

r4mmer: Did not mean to include this so I removed this change

tuliomir (approved):

BrunoCampana (approved):

PR #533, opened 2025-03-11
feat: use xpub stream as default

r4mmer:
### Acceptance Criteria
- Change the default history sync mode to use manual streaming.

### Security Checklist


pedroferreira1 (approved):

src/services/wallets.service.js
tuliomir, line 1: :star: An elegant solution for a lot of conditionals and fallbacks

tuliomir (approved):

PR #552, opened 2025-05-20
feat: pass contract_id to oracle signing method

r4mmer:
### Acceptance Criteria
- Add contract id to get oracle signed data api

### Security Checklist

### DIFF
--- a/__tests__/__fixtures__/http-fixtures.js
+++ b/__tests__/__fixtures__/http-fixtures.js
@@ -1020,8 +1020,7 @@ export default {
     success: true,
     tx: {
       hash: '00000008707722cde59ac9e7f4d44efbd3a5bd5f244223816ee676d328943b1b',
-      version: 4,
-      nc_blueprint_id: '3cb032600bdf7db784800e4ea911b10676fa2f67591f82bb62628c234e771595',
+      version: 0,
       nonce: '0',
       timestamp: 1572636346,
       weight: 1,
@@ -1065,7 +1064,7 @@ export default {
     },
     spent_outputs: {},
     meta: {
-      hash: '5c02adea056d7b43e83171a0e2d226d564c791d583b32e9a404ef53a2e1b363a',
+      hash: '00000008707722cde59ac9e7f4d44efbd3a5bd5f244223816ee676d328943b1b',
       spent_outputs: [],
       received_by: [],
       children: [],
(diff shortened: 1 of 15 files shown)


__tests__/__fixtures__/http-fixtures.js
```diff
@@ -1082,39 +1081,166 @@ export default {
...
+        ]
+      ],
+      nc_execution: 'success',
+      accumulated_weight: 17.441809467146022,
+      score: 0.0,
+      first_block_height: 19
+    },
+    spent_outputs: {}
+  },
+  '/transaction?id=0000045a5460cc1d00489c39ae4438c92d26180d996243d2dca8f7c4c62b7b50': {
```
r4mmer, line 1147: I'm also including a bet transaction, so it can be used in tests if needed

src/services/fireblocks.service.js
```diff
@@ -91,14 +91,24 @@ async function getTxSignatures(tx, storage, client) {
     });
   }
 
-  /* istanbul ignore next */
-  if (tx.version === hathorLib.constants.NANO_CONTRACTS_VERSION) {
+  /** @type {hathorLib.Address} */
+  let address = null;
```
r4mmer, line 95: These changes to fireblocks are due to the changes on the lib on how we sign the nano contract transactions

src/services/hsm.service.js
```diff
@@ -219,17 +219,33 @@ class HsmSession {
       });
     }
 
-    /* istanbul ignore next */
-    if (tx.version === hathorLib.constants.NANO_CONTRACTS_VERSION) {
+    /** @type {hathorLib.Address} */
+    let address = null;
```
r4mmer, line 223: Same changes as the fireblocks file, due to how we changed the signature for nano contracts on the lib

src/services/fireblocks.service.js
```diff
@@ -148,7 +158,19 @@ async function getTxSignatures(tx, storage, client) {
     if (!sig) {
       throw new Error(`Fireblocks did not return signature for address index ${ncCallerIndex} in transaction(${fbTxInfo.id}); external-id: (${dataToSignHash.toString('hex')})`);
     }
-    ncCallerSignature = fireblocksSignatureToDER(sig.signature);
+    const pubkeyHex = await storage.getAddressPubkey(ncCallerIndex);
+    if (sig.publicKey !== pubkeyHex) {
+      throw new Error(`Fireblocks signature does not match locally generated public key in transaction(${fbTxInfo.id}); external-id: (${dataToSignHash.toString('hex')})`);
+    }
```
pedroferreira1, line 164: suggestion(non-blocking): this could be moved inside the `if (tx.isNanoContract())`

pedroferreira1 (approved):

src/services/hsm.service.js
```diff
@@ -219,17 +219,33 @@ class HsmSession {
...
-      const address = hathorLib.addressUtils.getAddressFromPubkey(pubkey.toString('hex'), storage.config.getNetwork());
+      address = hathorLib.addressUtils.getAddressFromPubkey(pubkey.toString('hex'), storage.config.getNetwork());
+    }
+
+    if (address) {
       const addressInfo = await storage.getAddressInfo(address.base58);
       if (!addressInfo) {
-        // Not a wallet address
-        return { inputSignatures };
+        throw new Error(`No address info found for ${address.base58}`);
```
tuliomir, line 237: question: Is that a breaking change from the expected behavior? Are there any cases that would expect a valid response with the `inputSignatures` that will now break instead?

tuliomir (approved):

src/services/fireblocks.service.js
r4mmer: Since this validation works for both the if and else blocks I think it should remain outside the if. what do you think.

pedroferreira1: It's ok. You don't use `sig.publicKey` but you can validate it.

